PR #2, opened 2016-03-12
Add tracking for time spent in Present

jenatali:
Gives a rough calculation of how much time an app spent on the CPU producing work, assuming that all throttling is done via naively hitting the frame latency limits. If an app is self-throttling (e.g. a D3D12 app waiting on fences), or hits stalls due to things like mapping a resource, this simple metric won't give any data there.

This is preliminary for enabling capturing additional events which are emitted in the context of a DXGI present, but still provides new and interesting data.


punkUser:
Looks good! Just to confirm - do you agree to allow your contribution to be licensed by the project owners under the terms of the [included license](https://github.com/GameTechDev/PresentMon/blob/master/license.txt)?

PR #20, opened 2017-02-02
Fix an issue which can occur if a runtime present stop gets dropped, …

jenatali:
…or the runtime providers stop prior to kernel providers.

We saw some issues running PresentMon on ETL traces where DXGI's events stopped in the middle of a present, so multiple sets of kernel events were getting attributed to the same app present, resulting in multiple-completion of the frame.


jdm3:
Hi Jesse, I'm trying to understand the fix in the context of all the refactoring.  Is the fix wholly contained in c52c02f9ea724c7e05cbc3ebf16d0f3c9a065ba5?

jenatali:
Whoops… yeah, the fix is wholly contained c52c02f. Sorry, I’m still a relative novice at git branching. The refactoring is just something that we wanted to do in order to re-use the code without having to fork it.

-Jesse

jdm3:
No problem; I'm just trying to tease it apart to understand it.  The refactoring looks good too though.

Can I ask that you add copyright/license information to the new files that you've created?  For the cases where you moved existing code into a new file please reproduce the header from the original files.  For files containing only new code you can add your own copyright/license.

PR #27, opened 2017-05-06
Fix presents getting stuck in PresentMon.

jenatali:
A once-and-for-all fix for not wanting presents to get 'stuck' in the present-by-thread map if we miss the critical event that would normally remove it from that map.

Previously, the attempt to fix this broke certain types of presents. To fix it without regressions, each consumer of FindOrCreatePresent will check to ensure that the found event meets its expectations. Most call sites should have 'brand new' presents with no previous information except maybe a runtime's PresentStart, except for FlipMPO (which can see a previous FlipMPO event for the same frame) and SubmitPresentHistory (which can see a Blt beforehand). Any other information stored in the present event will cause it to be discarded as likely 'stuck' and created anew.

With this, it's likely safe to start relaxing the lost events logic, as we should be much more robust to it. Data generated while events are getting dropped might not be accurate, but if events stop getting dropped we should recover. Once one present from a swapchain/process completes, all the previous ones will get flushed as 'unknown' completion state.

### DIFF
--- a/source/PresentMonTraceConsumer.cpp
+++ b/source/PresentMonTraceConsumer.cpp
@@ -111,6 +111,14 @@ void PMTraceConsumer::OnDXGKrnlEvent(PEVENT_RECORD pEventRecord)
         // to trace the flip to screen.
         auto eventIter = FindOrCreatePresent(hdr);
 
+        // Check if we might have retrieved a 'stuck' present from a previous frame.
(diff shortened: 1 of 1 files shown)


jdm3:
Thanks Jesse!  Moved this to v1.1.0 branch, included in v1.1.1 release, and merged into master.

PR #28, opened 2017-05-07
Move hWnd association for GPU-GDI blt model presents to token propaga…

jenatali:
…tion instead of creation. This aligns to when DWM receives the token and decides which frame should be composed into a window, and is where similar logic already existed for composition buffer presents. This was causing PresentMon's latencies for this mode to be off by ~1 VSync for some apps.

Also fixed several regressions in ETL file processing introduced by refactoring:
1. ETL file is not mutually exclusive with filename.
2. ETL files can cause the ETW thread to exit without an external signal. This was triggering asserts. When this happens, the app should exit cleanly (at least, that's how it used to behave).
3. The start time for a trace session was uninitialized and caused CSV files to have impossible "time in second" values for each frame.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,10 @@ SOFTWARE.
 ## Command line options
 
 ```html
-Capture target options (use one of the following):
+Capture target options:
     -captureall                Record all processes (default).
     -process_name [exe name]   Record specific process specified by name.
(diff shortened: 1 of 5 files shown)


jdm3:
Thanks Jesse!  Moved this to v1.1.0 branch, included in v1.1.1 release, and merged into master.

PR #29, opened 2017-05-07
Add a -verbose mode which adds additional columns. Currently the only…

jenatali:
… column is a "present batching" column, which indicates whether or not the driver is batching presents.

Internally merged with -simple using a verbosity enum.

I've been sitting on this change for months since I wasn't sure the concept of present batching would be useful outside of the Windows graphics team for debugging driver issues, but some conversations with coworkers convinced me it could be. I manually ported it over after all the refactoring that's been done and polished it up by putting it behind a -verbose option.


jdm3:
Cherry-picked onto master.

Issue #55, opened 2018-11-30
Consume ETW metadata from within trace files


jenatali:
When a trace is collected and run through XPerf's -merge functionality (which is done for GPUView traces), there's additional synthetic events that are injected just before the first instance of each event that is parseable on the machine doing the merge, which contains all of the metadata necessary to decode the payload of these events.

Ideally, PresentMon should be able to consume these instead of going to TDH, in order to support capturing a trace on one system, and running PresentMon on it on another system, without having to worry about the event format changing between the two.

JeffersonMontgomery-Intel:
Merged into master, should generate a release soon just hoping to get a few more small changes in.

PR #56, opened 2018-12-17
Add support for consuming XPerf's metadata provider.

jenatali:
Implements #55 

This provider consists purely of events that were injected into ETL files during the merge process. On the machine doing the merge, TDH is queried and the resulting metadata is included in the stream, so events can be decoded on machines which don't have the same version of ETW manifests as the one which produced the ETL.

Windows Performance Toolkit comes with a KernelTraceControl.h containing the definition of the metadata provider GUID and the opcode value for the metadata events. These events payload a TRACE_EVENT_INFO struct for some other provider/event.

Unfortunately, TDH doesn't have APIs for decoding based on a provided TRACE_EVENT_INFO, only ones which operate on an internally-located TRACE_EVENT_INFO from (e.g.) an XML manifest. So the payload parsing has to be done manually. Currently, PresentMon only queries data from simple (no structs or arrays) events, so the parsing is kept simple.

For the MR data, all of the events are tracelogging, which contain their metadata within the event itself, so TDH can be used to parse those safely no matter what machine is doing the decoding.

Also... update the SDK version to the latest, since my home PC doesn't have the older ones. I'm open to reverting this portion if there's objections.

### DIFF
--- a/PresentData/KernelTraceControl.h
+++ b/PresentData/KernelTraceControl.h
@@ -1,0 +1,182 @@
+/*++
+
+Copyright (c) Microsoft Corporation.  All rights reserved.
+
+Module Name:
+
+    KernelTraceControl.h
+
+Abstract:
+
+    Public headers for event tracing control applications,
+    consumers and providers
+
+--*/
+
+#ifndef _KERNELTRACECONTROL_H_
+#define _KERNELTRACECONTROL_H_
+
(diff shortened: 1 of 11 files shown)


jenatali:
@JeffersonMontgomery-Intel would you be able to take a look at this? It's been sitting for >1 month. Thanks!

PR #75, opened 2020-03-27
Add CompositionSurfaceLuid and SwapChain Size to PresentEvent Struct

ZuomingShi:
Adding these SwainChian size should allow PresentData users to detect swapchain resize events, while adding CompositionSurfaceLuid allows users to correlate presentation events with other ETW events on Windows.


PresentData/PresentMonTraceConsumer.hpp
```diff
@@ -97,6 +97,9 @@ struct PresentEvent {
     Runtime Runtime;
     PresentMode PresentMode;
     PresentResult FinalState;
+    uint32_t DestWidth;
```
jenatali: These new fields are uninitialized. Add them to the initializer list in PresentMonTraceConsumer.cpp.

ZuomingShi: Added, thanks!

PR #76, opened 2020-04-06
Add basic PresentMon CSV Snapshot Comparison Test

ZuomingShi:
Add basic PresentMon test that compares the csv output of PresentMon with a snapshot from the verious version, agnostic to column ordering and whitespace changes. Add a basic test case for D3D12 flip mode.

The test is compiled in a new project that uses the C++ GTest framework. It runs the binary produced by the PresentMon project to generate a csv from a pre-recorded .ETL file, and compare the output of that file against the output of a previous version. By default a process name to track must be specified for each test case.

If the csv files do not match, the test will fail and report the first instance of mismatch.

.csv files have been removed from .gitignore file to simplify the process of adding future test cases.


JeffersonMontgomery-Intel:
I have googletests in a different fork I'd like to align these with.  Is the TaefTest implementation (and generalized PRESENTMON_EXPECTE... macros) important for you or is it ok if I drop these when merging the two?

ZuomingShi:
Hi Jefferson, thanks for the suggestion! I chatted with some folks and we agreed that dropping TaefTest here is fine.

Were you thinking of starting a different PR?

JeffersonMontgomery-Intel:
No, I'll just take your commit here and add some changes on top to make it align better with some other stuff I have.  I'll post here so you can review the changes first.

ZuomingShi:
Sounds good, thanks!

ZuomingShi:
Hi @JeffersonMontgomery-Intel, when do you think the new changes will become available?

JeffersonMontgomery-Intel:
Can you take a look at the test branch here (I couldn't figure out how to add it to the PR): https://github.com/GameTechDev/PresentMon/commits/tests

Let me know if that all looks ok and I'll merge it all in.  Or, we can discuss any changes needed here still.

p.s., your ETL actually exposed a crazy-rare bug already (https://github.com/GameTechDev/PresentMon/commit/ba0fdbdc9448b40ff0614bc3e87f4b37bc0d53ee) :) :)

ZuomingShi:
Sure! The change is pretty big and I need to finish some internal tasks, but I'll try to get back by the end of this week, if not with more feedback, then at least with a better ETA.

Glad that it's caught something already :) Awesome!

One quick thing I saw that is that there are duplicate entries from line 42-49 on the .sln file. Is it missing the Win32 configuration? https://github.com/GameTechDev/PresentMon/blob/tests/PresentMon.sln

ZuomingShi:
Sorry, I got sidetracked and didn't get a chance to look at it this week. I'll be prioritizing looking at it next week.

ZuomingShi:
Hi Jefferson, the PR looks fine to me, thanks! Only the change (cosmetic) I mentioned earlier comes to mind: 

The .sln file has some duplicate entires in line 42-49 https://github.com/GameTechDev/PresentMon/blob/tests/PresentMon.sln

Feel free to merge the tests branch into master whenever.

JeffersonMontgomery-Intel:
I think this is all in master now.

I decided to provide the ETL/CSV pairs through a release binary instead of LFS checkins.  That may end up being more cumbersome, but after reading a bit about github/LFS restrictions I think this way gives us more control especially if we don't need to version the ETL/CSVs with PresentMon.

I added some more ETLs, so coverage is looking a little better. https://github.com/GameTechDev/PresentMon/tree/master/Tests#presentmontestetls-coverage  Main thing missing is some win7 examples, which I'll add next.

PR #81, opened 2020-05-26
Provide -shutdown option to terminate already running PresentMon

pskhodad:
To terminate already running PresentMon process user can launch PresentMon with -shutdown option.

### DIFF
--- a/PresentMon/CommandLine.cpp
+++ b/PresentMon/CommandLine.cpp
@@ -248,6 +248,7 @@ static void PrintHelp()
                                     " listed as '<error>'.",
         "-terminate_on_proc_exit",  "Terminate PresentMon when all the target processes have exited.",
         "-terminate_after_timed",   "When using -timed, terminate PresentMon after the timed capture completes.",
+        "-shutdown",                "Terminate already running PresentMon process",
 
         "Beta options", nullptr,
         "-include_mixed_reality",   "Capture Windows Mixed Reality data to a CSV file with \"_WMR\" suffix.",
@@ -325,6 +326,7 @@ bool ParseCommandLine(int argc, char** argv)
     args->mIncludeWindowsMixedReality = false;
     args->mMultiCsv = false;
     args->mStopExistingSession = false;
+    args->mShutdown = false;
 
     bool simple = false;
     bool verbose = false;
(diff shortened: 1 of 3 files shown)


JeffersonMontgomery-Intel:
Thank-you, looks like a useful feature!

Looking through this, my first thought was wondering if this will work well if there are multiple PresentMon processes running.  Then I realized we already have this capability, albeit in a cumbersome way :).   The following will terminate any existing PresentMon process that is using the default session name (and you can target a particular process using `-session_name` as well):
```
PresentMon-dev-x64.exe -no_csv -dont_restart_as_admin -stop_existing_session -timed 0 -terminate_after_timed
```

So, I'm curious if that's enough?  Maybe we can just add a command line argument that is essentially an alias to that (plus maybe with some output cleanup)... something like:
```
-terminate_existing [session_name]
```

pskhodad:
Agree if this can be achieved with existing command line options then its better alternative

`PresentMon-dev-x64.exe -no_csv -dont_restart_as_admin -stop_existing_session -timed 0 -terminate_after_timed`

I had mixed result with above command-line, target PresentMon process exited but then new one continued to run on few occasions. Something time sensitive thing at play.

JeffersonMontgomery-Intel:
Hmm.  Maybe there's a race or bug with `-timed 0 -terminate_after_timed`... -timed 0 isn't a common usage.  I'll try to look into that.  Is there anything specific you noticed about when it continues running compared to when it stops?  How often does it continue running vs not?

pskhodad:
> How often does it continue running vs not?

1/10  times it continued to run, although repro was pretty consistent if I launch PresentMon that does the shutdown job using debugger. Initially I had breakpoint set between stop and start of session but later on it was happening even without breakpoint disabled. 

>         status = TraceSession::StopNamedSession(args.mSessionName);
>         if (status == ERROR_SUCCESS) {
>             status = gSession.Start(gPMConsumer, gMRConsumer, args.mEtlFileName, args.mSessionName);
>         }
> 

> Is there anything specific you noticed about when it continues running compared to when it stops?

Although not difference but common between two is - RECORDING message to console is always displayed even when it successfully shutdowns first instance and terminates itself.

JeffersonMontgomery-Intel:
OK, I think I've found the issue... will work on a fix and look at adding `-terminate_existing` at the same time.

JeffersonMontgomery-Intel:
Master now has command line argument that should implement this:

```
  -terminate_existing       Terminate any existing PresentMon realtime trace
                            sessions, then exit.  Use with -session_name to target
                            particular sessions.
```

Let me know if you see any issues!

PR #88, opened 2020-10-14
Allow PresentEvents to be constructed and removed like a PoD type

Zu-shi:
PresentData's destructor has asserts to catch incomplete events being destroyed. While this can be useful scenarios, it creates unexpected behaviors when these events are copied for diagnostic or record-keeping purposes.

My perspective is that this assert is better done in the debugger to keep the code clean and easy to use.

Alternatively, additional flags could help as a workaround, but isn't as clean.

### DIFF
--- a/PresentData/MixedRealityTraceConsumer.cpp
+++ b/PresentData/MixedRealityTraceConsumer.cpp
@@ -122,11 +122,6 @@ LateStageReprojectionEvent::LateStageReprojectionEvent(EVENT_HEADER const& hdr)
 {
 }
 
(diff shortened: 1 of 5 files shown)


JeffersonMontgomery-Intel:
Can you move the assert to here instead of just removing it?

https://github.com/GameTechDev/PresentMon/blob/main/PresentMon/OutputThread.cpp#L310

Zu-shi:
Makes sense, just added, and added the same logic to Late Stage Reprojection frames.

PR #89, opened 2020-11-04
Store PresentEvents in a Circular Buffer to avoid potential leaks

Zu-shi:
This changes adds mAllPresents to track all Present Events. This struct is a circular buffer and will remove and report events that weren't complete by the time the index wraps around.

To ensure that all shared pointers of the removed events are gone, RemovePresentFromTemporaryTrackingCollections is added to clean up presents at any of its lifetime stages. CompletePresent has been modified to use this function, and PresentEvent has a few additional tracking variables to facilitate fast and correct removal.

Some present event cleanup code scattered throughout PMTraceConsumer has been modified to work with CompletePresent, and some comments around ambiguity in the code has been added for future resolution.

Signed-off-by: Zuoming Shi <ZuomingShi@gmail.com>


PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1197,6 +1395,11 @@ void PMTraceConsumer::CompletePresent(std::shared_ptr<PresentEvent> p, uint32_t
     if (*presentIter == p) {
         std::lock_guard<std::mutex> lock(mPresentEventMutex);
         while (presentIter != presentDeque.end() && presentIter->get()->Completed) {
+            // TODO: we have to choose between pushing presentIter vs presentDeque[0] here. presentDeque[0] is likely the more reliable one.
```
jenatali, line 1398: I'd move the assert to replace the comment, and I'd try to be more consistent. `mPresentEvents` is modified using `*presentIter` while `mAllPresents` is modified using `presentDeque[0]`. Those should be the same, regardless of which one it is. In fact, `presentIter` could entirely be deleted and all accesses could be replaced with `presentDeque[0]`, or my preference, `presentDeque.front()`.

I think that'd increase the readability of the below code where we always `pop_front()` even though we have an iterator.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    // mPresentsByProcessAndSwapChain and mPresentsByProcess should always track the present's lifetime,
+    // so these also have an assert to validate this assumption.
+
+    DebugLostPresent(p);
+
+    // Only incomplete presents should make it here.
+    assert(!p->Completed);
+
+    // Remove the present from any struct that would only host the event temporarily.
+    // Should we loop through and remove the dependent presents?
```
jenatali, line 1314: This probably doesn't belong as a code comment. Should we? My gut says yes.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    RemovePresentFromTemporaryTrackingCollections(p);
+
+    // mPresentsByProcess
+    {
+        auto& presentsByThisProcess = mPresentsByProcess[p->ProcessId];
+        auto eventIter = presentsByThisProcess.find(p->QpcTime);
+        if (eventIter != presentsByThisProcess.end() && (eventIter->second == p)) {
+            presentsByThisProcess.erase(eventIter);
+            // TODO: Assert elementsRemoved = 1 if this element must exist. (What if QpcTime is not unique enough?)
```
jenatali, line 1323: This seems like a good idea to me. Maybe just call the overload of `erase` that takes the key directly instead of doing a find first, and assert that the return value is <= 1.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+        }
+    }
+
+    // mPresentsByProcessAndSwapChain
+    {
+        auto& presentDeque = mPresentsByProcessAndSwapChain[std::make_tuple(p->ProcessId, p->SwapChainAddress)];
+        auto presentIter = presentDeque.begin();
```
jenatali, line 1330: Personal preference, I'd write this as either a `for` loop:
```
for (auto presentIter = presentDeque.begin(); presentIter != presentDeque.end(); presentIter++)
{
    if (p == *presentIter)
   {
       presentDeque.erase(presentIter);
       break;
   }
}
```

Or, as a `std::find`:
```
auto presentIter = std::find(presentDeque.begin(), presentDeque.end(), p);
assert(presentIter != presentDeque.end());
presentDeque.erase(presentIter);
```

A while loop where the only thing in the body is an increment seems wrong.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+            // If we are in this loop for dozens of times, something is likely wrong.
+            presentIter++;
+        }
+        assert(presentIter != presentDeque.end());
+        presentDeque.erase(presentIter);
+    }
+
+    // Update the count of lost presents.
+    auto processAndThreadKey = std::make_tuple(
+        mAllPresents.at(mCurrentPresentTrackingIndex)->ProcessId,
```
jenatali, line 1343: Why not just use `p` to get at the process ID and thread ID?

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+
+    // Update the count of lost presents.
+    auto processAndThreadKey = std::make_tuple(
+        mAllPresents.at(mCurrentPresentTrackingIndex)->ProcessId, 
+        mAllPresents.at(mCurrentPresentTrackingIndex)->ThreadId
+    );
+    mCountLostPresentsByProcessAndSwapChain[processAndThreadKey]++;
+
+    // mAllPresents
+    {
```
jenatali, line 1349: Nitpick: Unnecessary extra scope.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+        auto eventIter = mPresentByThreadId.find(p->ThreadId);
+        if (eventIter != mPresentByThreadId.end() && eventIter->second == p) {
+            mPresentByThreadId.erase(eventIter);
+        }
+        // TODO: Assert that an element has been removed if this element must exist.
+    }
+
+    // mPresentsBySubmitSequence
+    {
+        // TODO: is 0 a valid QueueSubmitSequence?
```
jenatali, line 1208: No, you'll never have a valid submit sequence that's 0.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    // Currently defined as all structures except for mPresentsByProcess, 
+    // mPresentsByProcessAndSwapChain, and mAllPresents.
+
+    // mPresentByThreadId
+    {
+        auto eventIter = mPresentByThreadId.find(p->ThreadId);
+        if (eventIter != mPresentByThreadId.end() && eventIter->second == p) {
+            mPresentByThreadId.erase(eventIter);
+        }
+        // TODO: Assert that an element has been removed if this element must exist.
```
jenatali, line 1203: I'd delete this TODO, the condition to try to figure out if it should've been in this tracking struct is probably too complex to be worth it.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    }
+
+    // mPresentsBySubmitSequence
+    {
```
jenatali, line 1207: This function is full of unnecessary scopes. I'd remove them. A scope is great when it means local variables can't leak, or to force destruction to happen earlier, but a scope around an `if` which is going to introduce its own new scope isn't useful.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+        }
+    }
+
+    // mPresentsWaitingForDWM
+    {
+        if (p->WaitingForDwmCompletion)
+        {
+            auto eventIter = mPresentsWaitingForDWM.begin();
+            while (eventIter != mPresentsWaitingForDWM.end() && p != *eventIter)
+            {
```
jenatali, line 1275: Nitpick: Style inconsistency. Most of the code here puts open braces on the same line. Let's stick with that.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1265,6 +1468,20 @@ decltype(PMTraceConsumer::mPresentByThreadId.begin()) PMTraceConsumer::CreatePre
 {
     DebugCreatePresent(*newEvent);
 
+    newEvent->TrackingIndex = mCurrentPresentTrackingIndex;
+
+    if (mAllPresents[mCurrentPresentTrackingIndex] != nullptr)
+    {
+        // Existing present not completed by the time the circular buffer has come around.
+        // Remove this lost present.
+        fprintf(stderr, "Warning: present from process %u overwritten.\n", mAllPresents[mCurrentPresentTrackingIndex]->ProcessId);
```
jenatali, line 1477: I think we don't always want this logging. This probably should be scoped in some way.

PresentData/PresentMonTraceConsumer.cpp
jenatali: And I'd delete the rest of the TODOs that say the same thing. I don't think they're worth it.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+            if (eventIter != mPresentsBySubmitSequence.end() && (eventIter->second == p)) {
+                mPresentsBySubmitSequence.erase(eventIter);
+                // TODO: Assert elementsRemoved = 1 if this element must exist.
+            }
+        }
+    }
+
+    // mWin32KPresentHistoryTokens
+    {
+        if (p->ValidHistoryTokenKey) {
```
jenatali, line 1220: You can just delete this and go based on the CompositionSurfaceLuid being nonzero.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+            if (eventIter != mDxgKrnlPresentHistoryTokens.end() && eventIter->second == p) {
+                mDxgKrnlPresentHistoryTokens.erase(eventIter);
+                // TODO: Assert that an element has been removed if this element must exist.
+            }
+        }
+    }
+
+    // mBltsByDxgContext
+    {
+        if (p->ValidHContext) {
```
jenatali, line 1248: Same, HContexts will always be nonzero.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,6 +1188,169 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+            // There's no hard guaruntee that this element must exist, no assert needed here.
+        }
+    }
+
+    // mPresentsWaitingForDWM
+    {
+        if (p->WaitingForDwmCompletion)
+        {
+            auto eventIter = mPresentsWaitingForDWM.begin();
+            while (eventIter != mPresentsWaitingForDWM.end() && p != *eventIter)
```
jenatali, line 1274: Same comment as below, this should really be a `std::find()`.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -949,7 +991,13 @@ void PMTraceConsumer::HandleWin32kEvent(EVENT_RECORD* pEventRecord)
         }
 
         PMTraceConsumer::Win32KPresentHistoryTokenKey key(CompositionSurfaceLuid, PresentCount, BindId);
+        // TODO: what if there is already an element in mWin32KPresentHistoryTokens?
```
jenatali, line 994: There won't be, since PresentCount is part of the key. If you add an assert, make sure that it doesn't impact retail code (i.e. lookups should be in the `assert()` expression or under `#ifdef`).

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -644,20 +680,27 @@ void PMTraceConsumer::HandleDXGKEvent(EVENT_RECORD* pEventRecord)
...
-            CompletePresent(eventIter->second);
-        }
+        // Create a temporary copy of the shared_ptr since we may erase the iterator before we are done with its data.
+        std::shared_ptr<PresentEvent> event = eventIter->second;
 
-        if (eventIter->second->ThreadId != hdr.ThreadId) {
-            if (eventIter->second->TimeTaken == 0) {
-                eventIter->second->TimeTaken = hdr.TimeStamp.QuadPart - eventIter->second->QpcTime;
+        if (event->ThreadId != hdr.ThreadId) {
+            // In this case the same presentEvent may have two entries in mPresentByThreadId. Remove the extra one.
```
jenatali, line 687: Are you sure about this? I don't think that's true.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -515,11 +547,15 @@ void PMTraceConsumer::HandleDxgkPropagatePresentHistoryEventArgs(EVENT_HEADER co
...
         (eventIter->second->PresentMode == PresentMode::Composed_Flip && !eventIter->second->SeenWin32KEvents)) {
         mPresentsWaitingForDWM.emplace_back(eventIter->second);
+        eventIter->second->WaitingForDwmCompletion = true;
     }
 
     if (eventIter->second->PresentMode == PresentMode::Composed_Copy_GPU_GDI) {
         // Manipulate the map here
         // When DWM is ready to present, we'll query for the most recent blt targeting this window and take it out of the map
+
+        // TODO: what if there is already an element in mLastWindowPresent?
```
jenatali, line 557: Yes, there could've been one here. My intention when writing this was that any that got "lost" here should only remain in temporary structures that were manipulated by `CompletePresent`, and would be removed once one blt frame eventually completed -- that'd cause all the previous ones to also be completed and marked as Discarded, and cleaned up at that point.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -490,11 +519,14 @@ void PMTraceConsumer::HandleDxgkSubmitPresentHistoryEventArgs(
         if (tokenData == 0) {
             // This is the best we can do, we won't be able to tell how many frames are actually displayed.
             mPresentsWaitingForDWM.emplace_back(eventIter->second);
+            eventIter->second->WaitingForDwmCompletion = true;
         } else {
+            // TODO: what if there is already an element in mPresentsByLegacyBlitToken?
```
jenatali, line 524: Tokens should be unique, so an assert would be a good idea here.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -471,6 +496,10 @@ void PMTraceConsumer::HandleDxgkSubmitPresentHistoryEventArgs(
     eventIter->second->FinalState = PresentResult::Unknown;
     eventIter->second->TokenPtr = token;
 
+    // TODO: What happens if there already is an item in mDxgKrnlPresentHistoryTokens?
```
jenatali, line 499: This token is a kernel memory address, and so should be unique unless it was deallocated/reallocated with the same pointer, while we had a frame stuck in this map. That should never happen, so an assert seems like a good idea here.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -316,11 +335,17 @@ void PMTraceConsumer::HandleDxgkQueueSubmit(
         TRACK_PRESENT_PATH(eventIter->second);
         DebugModifyPresent(*eventIter->second);
 
+        // TODO: What happens if there already is an item in mPresentsBySubmitSequence?
+        // We should assert if that's impossible, or remove the existing event from other tracking lists.
         eventIter->second->QueueSubmitSequence = submitSequence;
         mPresentsBySubmitSequence.emplace(submitSequence, eventIter->second);
 
         if (eventIter->second->PresentMode == PresentMode::Hardware_Legacy_Copy_To_Front_Buffer && !supportsDxgkPresentEvent) {
+            // TODO: What happens if there already is an item in mBltsByDxgContext?
```
jenatali, line 344: It's impossible. Look at the top of this function, if there was one there, it was already removed.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -316,11 +335,17 @@ void PMTraceConsumer::HandleDxgkQueueSubmit(
         TRACK_PRESENT_PATH(eventIter->second);
         DebugModifyPresent(*eventIter->second);
 
+        // TODO: What happens if there already is an item in mPresentsBySubmitSequence?
```
jenatali, line 338: Like I said below, submit sequences are unique.

jenatali (changes requested):

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Awesome, thanks for the clarification! Forgot to remove this comment in the last revision, will address in the next one.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Good point, now that I look at this function again it has a lot more if conditions than it started out with, will address in the next revision.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Not sure about this one, that was my inference from reading this case block, but I could well have misread.

I think the thing that trips me up here is that we look up the event in mPresentByThreadId by hdr.ThreadId , but expect some cases where event->ThreadId.

This should only be possible if the ThreadId of a present in mPresentByThreadId is modified without changing the key, or if a present has its key in mPresentByThreadId modified without changing its ThreadId. Is there somewhere in the code where we do this?

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Yeah, leaning towards it. Will add in the next revision.

PresentData/PresentMonTraceConsumer.cpp
jenatali: Oh, it can move to a different thread ID in the map, but this isn't the right place to put logic related to that.

A single present can be inserted into the map with two different thread IDs if:
1. `CreatePresent` is called directly by a D3D9/DXGI event. The thread ID stored in the present is this one, which is the thread the app called on.
2. If the present was batched, then `FindOrCreatePresent` will also add it to the map, keyed by the thread ID that the driver called into the kernel. This isn't currently stored in the present structure.

The `RuntimePresentStop` logic will remove that first entry, and the second one will be removed once we don't expect anything that's correlated by thread for the present anymore. We probably want to add a second thread ID to the present struct, which is assigned by `FindOrCreatePresent` when it gets added to the map a second time, so that can be cleaned up generically.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Just added, let me know if anything looks incorrect.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1291,6 +1278,12 @@ void PMTraceConsumer::RemoveLostPresent(std::shared_ptr<PresentEvent> p)
 
     DebugLostPresent(p);
 
+    // Presents dependent on this event can no longer be trakced.
+    auto dependentPresents = p->DependentPresents;
+    for (auto dependentIter = dependentPresents.begin(); dependentIter != dependentPresents.end(); dependentIter++) {
```
jenatali, line 1283: Suggestion:
```c++
for (auto& dependentPresent : dependentPresents) {
    RemoveLostPresent(dependentPresent)
}
```

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1191,94 +1188,84 @@ void PMTraceConsumer::RemovePresentFromTemporaryTrackingCollections(std::shared_
     // mPresentsByProcessAndSwapChain, and mAllPresents.
 
     // mPresentByThreadId
+    auto threadEventIter = mPresentByThreadId.find(p->ThreadId);
+    if (threadEventIter != mPresentByThreadId.end() && threadEventIter->second == p) {
+        mPresentByThreadId.erase(threadEventIter);
+    }
+    if (p->DriverBatchThreadId != 0)
     {
```
jenatali, line 1196: Nitpick: Formatting.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1299,46 +1292,40 @@ void PMTraceConsumer::RemoveLostPresent(std::shared_ptr<PresentEvent> p)
     RemovePresentFromTemporaryTrackingCollections(p);
 
     // mPresentsByProcess
-    {
-        auto& presentsByThisProcess = mPresentsByProcess[p->ProcessId];
-        auto elementsRemoved = presentsByThisProcess.erase(p->QpcTime);
-        assert(elementsRemoved <= 1);
-    }
+    auto& presentsByThisProcess = mPresentsByProcess[p->ProcessId];
+    auto elementsRemoved = presentsByThisProcess.erase(p->QpcTime);
```
jenatali: Looks like you lost the assert here?

jenatali (approved):

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Ok, in the latest iteration I replaced WasBatched with the more informative DriverBatchThreadId (0 means not batched), and now remove based on this Id too.

I also noticed that FindOrCreatePresent removes from the mPresentsByProcess list, which was previously assumed to always contain all presents. I removed the assert that something has to be removed in RemoveLostPresent.

PresentData/PresentMonTraceConsumer.cpp
jenatali: Oh, I see your comment now that this was intentional. In that case, I'd remove the `elementsRemoved` variable.

PresentData/PresentMonTraceConsumer.cpp
jenatali: Looks good.

PresentData/PresentMonTraceConsumer.cpp
jenatali: Looks good.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -271,6 +281,10 @@ void PMTraceConsumer::HandleDxgkFlip(EVENT_HEADER const& hdr, int32_t flipInterv
 
     // If this is the DWM thread, piggyback these pending presents on our fullscreen present
     if (hdr.ThreadId == DwmPresentThreadId) {
+        for (auto iter = mPresentsWaitingForDWM.begin(); iter != mPresentsWaitingForDWM.end(); iter++)
+        {
+            iter->get()->PresentInDwmWaitingStruct = false;
```
Zu-shi, line 286: Note that this optimization was not present in the first iteration. We now only check mPresentsWaitingForDWM in RemoveLostPresents if the list hasn't been swapped out.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Kept these together since they were both about mPresentByThreadId

Zu-shi: I'm a bit of a toss up on this, but added the spacing

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Heads up that I hit asserting on this condition (SubtmitSequence key doesn't already exist) when selfhosting PresentMon over night. Unfortunately the debugger connection seems to have dropped somehow, so I don't have any other info.

image

Will continue selfhosting to see if I hit it again.

Zu-shi: Just hit this again. Not sure if it's related, but I was moving a menu item in Microsoft Todo Windows App when this hit. Does anything stand out?

[Link to dump](https://www.dropbox.com/s/d37w5rlosgo67zt/AssertOnTwoSubmitQueues.zip?dl=0)

Event we want to insert:
```
    [+0x000] QpcTime          : 0x7cac0f2bb8 [Type: unsigned __int64]
    [+0x008] ProcessId        : 0x641c [Type: unsigned int]
    [+0x00c] ThreadId         : 0x1d34 [Type: unsigned int]
    [+0x010] TimeTaken        : 0x0 [Type: unsigned __int64]
    [+0x018] ReadyTime        : 0x0 [Type: unsigned __int64]
    [+0x020] ScreenTime       : 0x0 [Type: unsigned __int64]
    [+0x028] SwapChainAddress : 0x0 [Type: unsigned __int64]
    [+0x030] SyncInterval     : -1 [Type: int]
    [+0x034] PresentFlags     : 0x0 [Type: unsigned int]
    [+0x038] Hwnd             : 0x0 [Type: unsigned __int64]
    [+0x040] TokenPtr         : 0xffffc2853a3a58c0 [Type: unsigned __int64]
    [+0x048] QueueSubmitSequence : 0x7 [Type: unsigned int]
    [+0x04c] Runtime          : Other (2) [Type: Runtime]
    [+0x050] PresentMode      : Composed_Composition_Atlas (7) [Type: PresentMode]
    [+0x054] FinalState       : Unknown (0) [Type: PresentResult]
    [+0x058] DestWidth        : 0x0 [Type: unsigned int]
    [+0x05c] DestHeight       : 0x0 [Type: unsigned int]
    [+0x060] CompositionSurfaceLuid : 0x0 [Type: unsigned __int64]
    [+0x068] SupportsTearing  : false [Type: bool]
    [+0x069] MMIO             : false [Type: bool]
    [+0x06a] SeenDxgkPresent  : false [Type: bool]
    [+0x06b] SeenWin32KEvents : false [Type: bool]
    [+0x06c] DriverBatchThreadId : 0x0 [Type: unsigned int]
    [+0x070] DwmNotified      : false [Type: bool]
    [+0x071] Completed        : false [Type: bool]
    [+0x078] DependentPresents : { size=0x0 } [Type: std::deque<std::shared_ptr<PresentEvent>,std::allocator<std::shared_ptr<PresentEvent> > >]
    [+0x0a0] TrackingIndex    : 0x1541 [Type: unsigned int]
    [+0x0a8] DxgKrnlHContext  : 0x0 [Type: unsigned __int64]
    [+0x0b0] Win32KPresentCount : 0x0 [Type: unsigned __int64]
    [+0x0b8] Win32KBindId     : 0x0 [Type: unsigned __int64]
    [+0x0c0] LegacyBlitTokenData : 0x0 [Type: unsigned __int64]
    [+0x0c8] PresentInDwmWaitingStruct : false [Type: bool]

```

mPresentsBySubmitSequence:
```
    [<Raw View>]     [Type: std::map<unsigned int,std::shared_ptr<PresentEvent>,std::less<unsigned int>,std::allocator<std::pair<unsigned int const ,std::shared_ptr<PresentEvent> > > >]
    [comparator]     : less [Type: std::_Compressed_pair<std::less<unsigned int>,std::_Compressed_pair<std::allocator<std::_Tree_node<std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >,void *> >,std::_Tree_val<std::_Tree_simple_types<std::pair<unsigned int const ,std::shared_ptr<PresentEvent> > > >,1>,1>]
    [allocator]      : allocator [Type: std::_Compressed_pair<std::allocator<std::_Tree_node<std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >,void *> >,std::_Tree_val<std::_Tree_simple_types<std::pair<unsigned int const ,std::shared_ptr<PresentEvent> > > >,1>]
    [0x0]            : 0x7, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x1]            : 0x59a, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x2]            : 0x59d, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x3]            : 0xbfcbc, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x4]            : 0xbfcc1, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x5]            : 0xbfcc6, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x6]            : 0x6eae1c, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
    [0x7]            : 0x6eae20, {...} [Type: std::pair<unsigned int const ,std::shared_ptr<PresentEvent> >]
```

Existing event: 
```
    [<Raw View>]     [Type: std::shared_ptr<PresentEvent>]
    [ptr]            : 0x1e140f88c60 [Type: PresentEvent *]
    [control block]  [Type: std::_Ref_count_obj2<PresentEvent> (derived from std::_Ref_count_base)]
    [+0x000] QpcTime          : 0x7cac0ddcbc [Type: unsigned __int64]
    [+0x008] ProcessId        : 0x55dc [Type: unsigned int]
    [+0x00c] ThreadId         : 0x222c [Type: unsigned int]
    [+0x010] TimeTaken        : 0x0 [Type: unsigned __int64]
    [+0x018] ReadyTime        : 0x7cac0de29c [Type: unsigned __int64]
    [+0x020] ScreenTime       : 0x0 [Type: unsigned __int64]
    [+0x028] SwapChainAddress : 0x0 [Type: unsigned __int64]
    [+0x030] SyncInterval     : -1 [Type: int]
    [+0x034] PresentFlags     : 0x0 [Type: unsigned int]
    [+0x038] Hwnd             : 0x0 [Type: unsigned __int64]
    [+0x040] TokenPtr         : 0xffffc2853a3a58c0 [Type: unsigned __int64]
    [+0x048] QueueSubmitSequence : 0x7 [Type: unsigned int]
    [+0x04c] Runtime          : Other (2) [Type: Runtime]
    [+0x050] PresentMode      : Composed_Composition_Atlas (7) [Type: PresentMode]
    [+0x054] FinalState       : Unknown (0) [Type: PresentResult]
    [+0x058] DestWidth        : 0x0 [Type: unsigned int]
    [+0x05c] DestHeight       : 0x0 [Type: unsigned int]
    [+0x060] CompositionSurfaceLuid : 0x0 [Type: unsigned __int64]
    [+0x068] SupportsTearing  : false [Type: bool]
    [+0x069] MMIO             : false [Type: bool]
    [+0x06a] SeenDxgkPresent  : true [Type: bool]
    [+0x06b] SeenWin32KEvents : false [Type: bool]
    [+0x06c] DriverBatchThreadId : 0x0 [Type: unsigned int]
    [+0x070] DwmNotified      : false [Type: bool]
    [+0x071] Completed        : false [Type: bool]
    [+0x078] DependentPresents : { size=0x0 } [Type: std::deque<std::shared_ptr<PresentEvent>,std::allocator<std::shared_ptr<PresentEvent> > >]
    [+0x0a0] TrackingIndex    : 0x153f [Type: unsigned int]
    [+0x0a8] DxgKrnlHContext  : 0x0 [Type: unsigned __int64]
    [+0x0b0] Win32KPresentCount : 0x0 [Type: unsigned __int64]
    [+0x0b8] Win32KBindId     : 0x0 [Type: unsigned __int64]
    [+0x0c0] LegacyBlitTokenData : 0x0 [Type: unsigned __int64]
    [+0x0c8] PresentInDwmWaitingStruct : false [Type: bool]
```

Zu-shi: Note: event we want to insert came from process 0x641c = Todo.exe,

Preexisting event came from process 0x55dc = DataExchangeHost.exe,

Zu-shi: It's also interesting that looking at the TrackingIndex, these presents were created one right after another.

jenatali: Huh... looks like the submit sequence is per-context, not global. Unfortunately, some events that use the submit sequence, such as MMIOFlip or the VSync DPCs, don't provide context information, so we can't just simply change the key to (context, submit sequence).

For full determinism, we'd need to split this container in two, one which is keyed on (context, submit sequence) for tracking rendering from submission to completion, and one which is keyed on (adapter, vidpn source/target ID, submit sequence) for tracking flips, I think.

This is a good find. Maybe let's just work around it for now? Though I guess I don't really have a great idea how to work around it.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -644,20 +677,25 @@ void PMTraceConsumer::HandleDXGKEvent(EVENT_RECORD* pEventRecord)
...
-            if (eventIter->second->TimeTaken == 0) {
-                eventIter->second->TimeTaken = hdr.TimeStamp.QuadPart - eventIter->second->QpcTime;
+        if (event->ThreadId != hdr.ThreadId) {
+            if (event->TimeTaken == 0) {
+                event->TimeTaken = hdr.TimeStamp.QuadPart - event->QpcTime;
             }
-            eventIter->second->WasBatched = true;
+            event->DriverBatchThreadId = hdr.ThreadId;
+
             mPresentByThreadId.erase(eventIter);
```
JeffersonMontgomery-Intel, line 689: eventITer->ThreadId = 0 (so that CompleteEvent() doesn't have to search for it)

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,12 +1181,152 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    // LegacyTokenData cannot be 0 if it's in mPresentsByLegacyBlitToken list.
+    if (p->LegacyBlitTokenData != 0) {
+        auto eventIter = mPresentsByLegacyBlitToken.find(p->LegacyBlitTokenData);
+        if (eventIter != mPresentsByLegacyBlitToken.end() && eventIter->second == p) {
+            mPresentsByLegacyBlitToken.erase(eventIter);
+        }
+    }
+}
+
+void PMTraceConsumer::RemoveLostPresent(std::shared_ptr<PresentEvent> p)
```
JeffersonMontgomery-Intel, line 1274: This looks similar enough to CompletePresent(), especially if we move to the similar modle of dequeing lost presents too.  I think we should combine these functions with a parameter for Complete vs Lost.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1139,12 +1181,152 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
...
+    // mPresentsByLegacyBlitToken
+    // LegacyTokenData cannot be 0 if it's in mPresentsByLegacyBlitToken list.
+    if (p->LegacyBlitTokenData != 0) {
+        auto eventIter = mPresentsByLegacyBlitToken.find(p->LegacyBlitTokenData);
+        if (eventIter != mPresentsByLegacyBlitToken.end() && eventIter->second == p) {
+            mPresentsByLegacyBlitToken.erase(eventIter);
+        }
+    }
+}
+
```
JeffersonMontgomery-Intel, line 1273: I don't see any interaction with the existing `HandleStuckPresent()`.  The old concept of a stuck present needs to be ported over... and I think that just means calling RemoveLostPresent() instead.

JeffersonMontgomery-Intel (changes requested):
This is looking pretty good!  I posted various comments, a bunch are minor/nitpicking and I can just handle them as part of the merge if needed.  But there's a few I hope you can look at (dequeueing lost presents, handling stuck presents, and keeping present keys consistent)?

Zu-shi:
> This is looking pretty good! I posted various comments, a bunch are minor/nitpicking and I can just handle them as part of the merge if needed. But there's a few I hope you can look at (dequeueing lost presents, handling stuck presents, and keeping present keys consistent)?

Sounds good. I was out yesterday but will take a pass through these today.

Best,
Z

PresentData/PresentMonTraceConsumer.hpp
```diff
@@ -103,12 +103,19 @@ struct PresentEvent {
...
     bool SeenWin32KEvents;
-    bool WasBatched;
+    uint32_t DriverBatchThreadId;
     bool DwmNotified;
     bool Completed;
 
-    // Additional transient state
+    // Additional transient tracking state
     std::deque<std::shared_ptr<PresentEvent>> DependentPresents;
+    uint32_t TrackingIndex;
```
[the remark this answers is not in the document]
Zu-shi: Sure to the comments part!

For the iterators, I think it's a righteous change but may be better in a future PR. It will be a pretty big incremental delta, and this PR has a lot of changes already currently.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -297,9 +312,15 @@ void PMTraceConsumer::HandleDxgkQueueSubmit(
...
                 if (eventIter->second->ScreenTime != 0) {
                     CompletePresent(eventIter->second);
+                    completedPresent = true;
                 }
             }
-            mBltsByDxgContext.erase(eventIter);
+
+            if (!completedPresent)
+            {   
+                mBltsByDxgContext.erase(eventIter);
```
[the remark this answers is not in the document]
Zu-shi: I think variables like DxgKrnlHContext means more "what properties we know about this present", as opposed to PresentInDwmWaitingStruct which are about "where this present was stored". So for variables like DxgKrnlHContext that give information about the presents, it may be more intuitive and easier to debug to the number is not reset.

In the future when iterators members are added to PresentEvents, I think those will help distinguish the role of DxgKrnlHContext better.

I agree that Add() and Remove() currently feel like overkill... it's a question of personal preferences though I'd lean towards not having them yet.

For ease tracking this convo, I'm going to resolve other similar comments about resetting variables. I can apply discussion results from this thread to the other comments as needed :)

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Replaced HandleStuckPresent() with RemoveLostPresent()

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: I think they are similar conceptually, but if merged may turn into two big if statements or several smaller if statements on the flag condition, which might not be a huge improvement over the current model. Orienting towards not merge them.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Heads up that this assert has triggered in selfhost:

https://www.dropbox.com/s/pc2b5rdifcbsrg9/TokenPtrDump.zip?dl=0

```
Prexisting: [Type: std::shared_ptr<PresentEvent>]
    [<Raw View>]     [Type: std::shared_ptr<PresentEvent>]
    [ptr]            : 0x1bb7e14d4b0 [Type: PresentEvent *]
    [control block]  [Type: std::_Ref_count_obj2<PresentEvent> (derived from std::_Ref_count_base)]
    [+0x000] QpcTime          : 0x99d3f4ecb0 [Type: unsigned __int64]
    [+0x008] ProcessId        : 0x76a4 [Type: unsigned int]
    [+0x00c] ThreadId         : 0x7a18 [Type: unsigned int]
    [+0x010] TimeTaken        : 0x25367 [Type: unsigned __int64]
    [+0x018] ReadyTime        : 0x0 [Type: unsigned __int64]
    [+0x020] ScreenTime       : 0x0 [Type: unsigned __int64]
    [+0x028] SwapChainAddress : 0x89b3140 [Type: unsigned __int64]
    [+0x030] SyncInterval     : -1 [Type: int]
    [+0x034] PresentFlags     : 0x0 [Type: unsigned int]
    [+0x038] Hwnd             : 0x381342 [Type: unsigned __int64]
    [+0x040] TokenPtr         : 0xffffe482a95d68c0 [Type: unsigned __int64]
    [+0x048] QueueSubmitSequence : 0x15ca4b [Type: unsigned int]
    [+0x04c] Runtime          : D3D9 (1) [Type: Runtime]
    [+0x050] PresentMode      : Composed_Copy_GPU_GDI (5) [Type: PresentMode]
    [+0x054] FinalState       : Unknown (0) [Type: PresentResult]
    [+0x058] DestWidth        : 0x0 [Type: unsigned int]
    [+0x05c] DestHeight       : 0x0 [Type: unsigned int]
    [+0x060] CompositionSurfaceLuid : 0x0 [Type: unsigned __int64]
    [+0x068] DriverBatchThreadId : 0x53d0 [Type: unsigned int]
    [+0x06c] SupportsTearing  : false [Type: bool]
    [+0x06d] MMIO             : false [Type: bool]
    [+0x06e] SeenDxgkPresent  : true [Type: bool]
    [+0x06f] SeenWin32KEvents : false [Type: bool]
    [+0x070] DwmNotified      : false [Type: bool]
    [+0x071] Completed        : false [Type: bool]
    [+0x072] IsLost           : false [Type: bool]
    [+0x074] mAllPresentsTrackingIndex : 0x1939 [Type: unsigned int]
    [+0x078] DxgKrnlHContext  : 0x0 [Type: unsigned __int64]
    [+0x080] Win32KPresentCount : 0x0 [Type: unsigned __int64]
    [+0x088] Win32KBindId     : 0x0 [Type: unsigned __int64]
    [+0x090] LegacyBlitTokenData : 0x0 [Type: unsigned __int64]
    [+0x098] DependentPresents : { size=0x0 } [Type: std::deque<std::shared_ptr<PresentEvent>,std::allocator<std::shared_ptr<PresentEvent> > >]
    [+0x0c0] PresentInDwmWaitingStruct : false [Type: bool]
```

New present: [Type: std::shared_ptr<PresentEvent>]
```
    [<Raw View>]     [Type: std::shared_ptr<PresentEvent>]
    [ptr]            : 0x1bb7e14fe50 [Type: PresentEvent *]
    [control block]  [Type: std::_Ref_count_obj2<PresentEvent> (derived from std::_Ref_count_base)]
    [+0x000] QpcTime          : 0x99d3f6ffca [Type: unsigned __int64]
    [+0x008] ProcessId        : 0x77ec [Type: unsigned int]
    [+0x00c] ThreadId         : 0xecc [Type: unsigned int]
    [+0x010] TimeTaken        : 0x10103db [Type: unsigned __int64]
    [+0x018] ReadyTime        : 0x0 [Type: unsigned __int64]
    [+0x020] ScreenTime       : 0x0 [Type: unsigned __int64]
    [+0x028] SwapChainAddress : 0x2021393d1f0 [Type: unsigned __int64]
    [+0x030] SyncInterval     : 1 [Type: int]
    [+0x034] PresentFlags     : 0x0 [Type: unsigned int]
    [+0x038] Hwnd             : 0x0 [Type: unsigned __int64]
    [+0x040] TokenPtr         : 0xffffe482a95d68c0 [Type: unsigned __int64]
    [+0x048] QueueSubmitSequence : 0x0 [Type: unsigned int]
    [+0x04c] Runtime          : DXGI (0) [Type: Runtime]
    [+0x050] PresentMode      : Composed_Flip (4) [Type: PresentMode]
    [+0x054] FinalState       : Unknown (0) [Type: PresentResult]
    [+0x058] DestWidth        : 0x384 [Type: unsigned int]
    [+0x05c] DestHeight       : 0x4ae [Type: unsigned int]
    [+0x060] CompositionSurfaceLuid : 0x9fceef2 [Type: unsigned __int64]
    [+0x068] DriverBatchThreadId : 0x0 [Type: unsigned int]
    [+0x06c] SupportsTearing  : false [Type: bool]
    [+0x06d] MMIO             : false [Type: bool]
    [+0x06e] SeenDxgkPresent  : false [Type: bool]
    [+0x06f] SeenWin32KEvents : true [Type: bool]
    [+0x070] DwmNotified      : false [Type: bool]
    [+0x071] Completed        : false [Type: bool]
    [+0x072] IsLost           : false [Type: bool]
    [+0x074] mAllPresentsTrackingIndex : 0x193a [Type: unsigned int]
    [+0x078] DxgKrnlHContext  : 0x0 [Type: unsigned __int64]
    [+0x080] Win32KPresentCount : 0x81e27 [Type: unsigned __int64]
    [+0x088] Win32KBindId     : 0x3 [Type: unsigned __int64]
    [+0x090] LegacyBlitTokenData : 0x0 [Type: unsigned __int64]
    [+0x098] DependentPresents : { size=0x0 } [Type: std::deque<std::shared_ptr<PresentEvent>,std::allocator<std::shared_ptr<PresentEvent> > >]
    [+0x0c0] PresentInDwmWaitingStruct : false [Type: bool]
```
PreExisting:
mAllPresentsTrackingIndex : 0x1939
30372 = Baba is You.exe

New:
mAllPresentsTrackingIndex : 0x193a
30700 = Solitaire.exe

Current mAllPresentsNextIndex: 0x193c

These are windowed blit and windowed flip apps on the same monitor in a multi-monitor scenario.

Neither has been marked as lost. Perhaps the old present got deallocated, stuck, and allocated again? If this is expected, then the assert might not be very useful if selfhost can hit it frequently.

Zu-shi: Screenshot:
image

Zu-shi: Also @jdm3 unfortunately I don't have a good way of being notified if you replied to a thread but didn't unresolve the conversation :( let me know if there are comments I accidentally ignored because they are in a resolved conversation.

jenatali: That's... odd. I have no idea how that would happen.

Also, there's no reason to put Win8+ flip model presents in that map, we only need Win7 flip, blt, and DComp (composition atlas) presents there. It probably doesn't hurt anything to put them there, and it might be extra complexity to avoid putting them in there, just wanted to call it out.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Heads up that replacing this in HandleDxgkBlt regressed Vulkan Windowed mode tracking, and replacing this in HandleDxgkFlip regressed Full Screen and Borderless Windowed Mode tracking. In both cases no presents are reported.

I think the issue here is that RuntimePresentStop isn't being called to remove these presents from mPresentByThreadId, so when the next  HandleDxgkBlt/HandleDxgkFlip call happens before the previous present has been completed (and thus removed from mPresentByThreadId), FindOrCreatePresent finds that previous present and thinks it's stuck.

Will likely solve this via keeping the HandleStuckPresent logic for Runtime::Other in my next PR.

@jenatali as FYI & check to see whether this seems right.

PresentData/PresentMonTraceConsumer.hpp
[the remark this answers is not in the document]
JeffersonMontgomery-Intel: OK

PresentData/PresentMonTraceConsumer.cpp
JeffersonMontgomery-Intel: That sounds right: in the VK case (or any without instrumented present API) it's essentially just a stuck "mPresentByThreadId" but we don't want to remove the whole present.

Is the right solution to not add to mPresentByThreadId if the present was created in HandleDxgkBlit/Flip()?

jenatali: That sounds like the right fix to me.

Zu-shi: Is the interaction with this part of FindOrCreatePresent a concern?
https://github.com/GameTechDev/PresentMon/blob/d1749c3f6711b27a4308eb1ff9f763238b5df08f/PresentData/PresentMonTraceConsumer.cpp#L1418

```
        // If not, lookup the oldest in-progress present that was created by
        // this process but still doesn't have a known PresentMode.  This is
        // the mechanism for DXGK/Win32K events for looking up a present
        // created by DXGI/D3D Present on a different thread. It assumes that
        // batched presents are popped off the front of the driver queue by
        // process in order.
```

I believes this means that the oldest in-progress present needs to have a known PresentMode in Vulkan by the point HandleDxgkBlt and HandleDxgkFlip calls FindOrCreatePresent. Is this safe to assume?

jenatali: In the case of Runtime::Other, there won't have been a present created yet. The Blt or Flip would be the first event you see in those cases.

Zu-shi: Not inserting the mPresentByThread in DxgBlit and DxgKrnl events didn't end up working, I think partially because Microsoft_Windows_DxgKrnl::Present_Info::Id: relies on mPresentsByThreadId to find it? This solution is a bit complex to reason about.

What about just removing non-D3D9/DXGI runtime presents from mPresentsByThreadId at the end of the Microsoft_Windows_DxgKrnl::Present_Info::Id case? This seems intuitive, and passes Vulkan tests without making exceptions for DxgBlit and DxgKrnl.

jenatali: That sound reasonable. You could even take it a step further and remove it from `mPresentsByThreadId` if either:
* The `Present_Info` event thread ID matches the driver thread ID
* The `Present_Info` event thread ID matches the primary thread ID, and the runtime is `Other`

In both of these cases, `Present_Info` is the last event we'd expect to see from the thread that called it.

Zu-shi: I think Microsoft_Windows_DxgKrnl::Present_Info::Id makes sense since mPresentsByThreadId's lifetime should correspond with as much as the runtime Present API lifetime as possible. In DXGI/D3D9 that would be RuntimePresentStop, in other runtimes, according to the comment, Present_Info should be the best stopping point.

```
    case Microsoft_Windows_DxgKrnl::Present_Info::Id:
    {
        // This event is emitted at the end of the kernel present, before returning.
        // The presence of this event is used with blt presents to indicate that no
        // PHT is to be expected.
```

Zu-shi: > That sound reasonable. You could even take it a step further and remove it from `mPresentsByThreadId` if either:
> 
> * The `Present_Info` event thread ID matches the driver thread ID
> * The `Present_Info` event thread ID matches the primary thread ID, and the runtime is `Other`
> 
> In both of these cases, `Present_Info` is the last event we'd expect to see from the thread that called it.

Interesting, does RuntimePresentStop not happen after Present_Info in the first case?

jenatali: It will (or already has), but not on the `Present_info` thread, it's on the same thread that did the DXGI/D3D9 start event.

Zu-shi: I went with https://github.com/GameTechDev/PresentMon/blob/41e50b9a86d43fe182bbf24cab0691b1f1f545f5/PresentData/PresentMonTraceConsumer.cpp#L700-L702

The thinking is that:
> The Present_Info event thread ID matches the driver thread ID
Is in this code block: https://github.com/GameTechDev/PresentMon/blob/41e50b9a86d43fe182bbf24cab0691b1f1f545f5/PresentData/PresentMonTraceConsumer.cpp#L689-L696

For 
> The Present_Info event thread ID matches the primary thread ID, and the runtime is Other

If runtime is Other, I think we won't get any other events to help remove the presents from mPresentByThreadId, so we should remove altogether.

Not sure if at any point you meant removing the DriverBatchThreadId member, for now they are probably not removed until the present completes.

PresentData/PresentMonTraceConsumer.hpp
```diff
@@ -320,10 +342,11 @@ struct PMTraceConsumer
 
     void CompletePresent(std::shared_ptr<PresentEvent> p, uint32_t recurseDepth=0);
     std::shared_ptr<PresentEvent> FindBySubmitSequence(uint32_t submitSequence);
-    decltype(mPresentByThreadId.begin()) FindOrCreatePresent(EVENT_HEADER const& hdr);
-    decltype(mPresentByThreadId.begin()) CreatePresent(std::shared_ptr<PresentEvent> present, decltype(mPresentsByProcess.begin()->second)& presentsByThisProcess);
+    std::shared_ptr<PresentEvent> FindOrCreatePresent(EVENT_HEADER const& hdr);
```
Zu-shi, line 345: The change in return value is not strictly related to the PR, just syntactic sugar that I added when working on an abandoned approach to restore Vulkan tracking. I decided to put it in the PR anyway but feel free to mention if there are concerns.

PresentData/PresentMonTraceConsumer.cpp
jenatali: That makes sense, I forgot about that block which removes it for batched presents.

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Resolving this one as out of scope for this PR.

JeffersonMontgomery-Intel (approved):
Thanks, this looks good.  I'll pull locally and do some testing then I can merge it in if you think it's ready?

Zu-shi:
> Thanks, this looks good. I'll pull locally and do some testing then I can merge it in if you think it's ready?

Yeah, I think this is ready!

PR #90, opened 2020-11-11
Add support for Blit Cancel events

jeffwick-msft:
There are cases where a Present Blit can be optimized out in kernel. In such cases, we return success to the caller, but issue no further work for the present, leaving PresentMon to see these as outstanding. A future OS release will issue a new ETW event when such a scenario occurs. This change detects this new event and marks the present as discarded.

Signed-off-by: Jeff Wickenheiser <jeffwick@microsoft.com>

### DIFF
--- a/PresentData/ETW/Microsoft_Windows_DxgKrnl.h
+++ b/PresentData/ETW/Microsoft_Windows_DxgKrnl.h
@@ -157,11 +157,6 @@ struct Blit_Info_Struct {
 };
 
 template<typename PointerT>
-struct Blit_Cancel_Struct {
-    PointerT    hwnd;
-};
-
-template<typename PointerT>
 struct Flip_Info_Struct {
     PointerT    pDmaBuffer;
     uint32_t    VidPnSourceId;
--- a/PresentData/ETW/Microsoft_Windows_DxgKrnl.h
+++ b/PresentData/ETW/Microsoft_Windows_DxgKrnl.h
@@ -111,6 +111,7 @@ enum class Channel : uint8_t {
 };
 
 EVENT_DESCRIPTOR_DECL(Blit_Info                     , 0x00a6, 0x00, 0x11, 0x04, 0x00, 0x0067, 0x4000000000000001)
+EVENT_DESCRIPTOR_DECL(Blit_Cancel                   , 0x01f5, 0x00, 0x11, 0x04, 0x00, 0x0135, 0x4000000000000001)
 EVENT_DESCRIPTOR_DECL(Flip_Info                     , 0x00a8, 0x00, 0x11, 0x00, 0x00, 0x0003, 0x4000000000000001)
 EVENT_DESCRIPTOR_DECL(FlipMultiPlaneOverlay_Info    , 0x00fc, 0x00, 0x11, 0x00, 0x00, 0x008f, 0x4000000000000001)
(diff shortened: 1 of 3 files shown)


PresentData/ETW/Microsoft_Windows_DxgKrnl.h
```diff
@@ -155,6 +156,11 @@ struct Blit_Info_Struct {
     uint32_t    SubRectCount;
 };
 
+template<typename PointerT>
+struct Blit_Cancel_Struct {
```
jenatali, line 160: I'm having trouble finding what these are actually used for. Do we need one here?

jenatali (approved):
LGTM.

PresentData/ETW/Microsoft_Windows_DxgKrnl.h
jeffwick-msft: We don't actually read any payload from this event, so this indeed goes unused. I can get rid of it.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -707,6 +717,12 @@ void PMTraceConsumer::HandleDXGKEvent(EVENT_RECORD* pEventRecord)
         HandleDxgkBlt(hdr, hwnd, bRedirectedPresent);
         break;
     }
+    case Microsoft_Windows_DxgKrnl::Blit_Cancel::Id:
+    {
+        TRACK_PRESENT_PATH_GENERATE_ID();
```
JeffersonMontgomery-Intel, line 722: This should be a TRACK_PRESENT_PATH() macro, instead of ..._GENERATE_ID().  I know it's sort of confusing... if there's any improvement to the comment on line 37 you can think of let me know :)

PresentData/ETW/Microsoft_Windows_DxgKrnl.h
JeffersonMontgomery-Intel: No they're generated by etw_list tool but not used by any PresentMon code.  That said, they can have use as documentation e.g., when re-running etw_list later it can be useful to see diffs in properties/versions etc.

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -235,6 +235,16 @@ void PMTraceConsumer::HandleDxgkBlt(EVENT_HEADER const& hdr, uint64_t hwnd, bool
     }
 }
 
+void PMTraceConsumer::HandleDxgkBltCancel(EVENT_HEADER const& hdr)
```
JeffersonMontgomery-Intel, line 238: Can you add a comment here (basically, your commit message)?

PresentData/ETW/Microsoft_Windows_DxgKrnl.h
jeffwick-msft: Is this one interesting for that case, given we don't utilize any of the data? Happy to add it back if you'd like it.

PresentData/PresentMonTraceConsumer.cpp
jeffwick-msft: Thanks, I hadn't seen the comment before... this makes more sense now.

PresentData/ETW/Microsoft_Windows_DxgKrnl.h
JeffersonMontgomery-Intel: Not a big deal... either way :)

PR #92, opened 2020-11-12
Add ability to scope data collection to specified process id

jeffwick-msft:
This change allows PresentMon data collection to be scoped to specified processes (currently only supporting process ids). This reduces memory usage by not tracking unnecessary presents from other processes, improving beyond the current implementation where filtering is just applied to the output.

Signed-off-by: Jeff Wickenheiser jeffwick@microsoft.com

### DIFF
--- a/PresentData/PresentMonTraceConsumer.cpp
+++ b/PresentData/PresentMonTraceConsumer.cpp
@@ -98,9 +98,10 @@ PresentEvent::PresentEvent(EVENT_HEADER const& hdr, ::Runtime runtime)
 #endif
 }
 
-PMTraceConsumer::PMTraceConsumer(bool filteredEvents, bool simple)
+PMTraceConsumer::PMTraceConsumer(bool filteredEvents, bool simple, bool trackedFiltering)
     : mFilteredEvents(filteredEvents)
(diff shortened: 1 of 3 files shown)


jenatali (approved):

Zu-shi (approved):

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -216,10 +227,19 @@ void PMTraceConsumer::HandleDxgkBlt(EVENT_HEADER const& hdr, uint64_t hwnd, bool
...
+
+    if (eventIter == mPresentByThreadId.end()) {
+        return;
+    }
+
     if (eventIter->second->PresentMode != PresentMode::Unknown) {
         HandleStuckPresent(hdr, &eventIter);
     }
 
+    if (eventIter == mPresentByThreadId.end()) {
```
JeffersonMontgomery-Intel, line 239: Can you move this (and few other instances below) into the scope that calls HandleStuckPresent()?

jeffwick-msft: Good feedback - done!

PR #93, opened 2020-11-14
Fix failures in Gold Csv caused by platform-dependent printf rounding

Zu-shi:
printf has inconsistent rounding behavior accross different platforms, see:
https://www.exploringbinary.com/inconsistent-rounding-of-printed-floating-point-numbers/

This PR adds some comparison code to tolerate rounding errors on the smallest decimal point.

Previously, tests cases 0 and 3 returns respectively:

GoldEtlCsvTests.cpp(110): error: Difference on line: 326
    COLUMN                    TEST VALUE                            GOLD VALUE
    MsUntilDisplayed          16.312                                  16.313

GoldEtlCsvTests.cpp(110): error: Difference on line: 119
    COLUMN                    TEST VALUE                            GOLD VALUE
    MsInPresentAPI             0.062                                    0.063

Now both tests pass. Validated via printing that this path is only stepped into in these two cases.

Signed-off-by: Zuoming Shi <ZuomingShi@gmail.com>

### DIFF
--- a/Tests/GoldEtlCsvTests.cpp
+++ b/Tests/GoldEtlCsvTests.cpp
@@ -112,15 +112,17 @@ public:
 
                     // Doubles should be good for 15 digits of precision.
 
-                    double testNumber = strtod(a, nullptr);
-                    double goldNumber = strtod(b, nullptr);
+                    double testNumber, goldNumber;
+                    int testSucceededCount = sscanf_s(a, "%lf", &testNumber);
+                    int goldSucceededCount = sscanf_s(b, "%lf", &goldNumber);
+
                     const char* testDecimalAddr = strchr(a, '.');
                     const char* goldDecimalAddr = strchr(b, '.');
-                    if (testNumber != 0.0 && goldNumber != 0.0 && testDecimalAddr != NULL && goldDecimalAddr != NULL) {
+                    if (testSucceededCount == 1 && goldSucceededCount == 1 && testDecimalAddr != NULL && goldDecimalAddr != NULL) {
                         size_t testDecimalNumbersCount = (a + strlen(a)) - testDecimalAddr - 1;
                         size_t goldDecimalNumbersCount = (b + strlen(b)) - goldDecimalAddr - 1;
                         double difference = (testNumber * pow(10.0, testDecimalNumbersCount)) - (goldNumber * pow(10.0, goldDecimalNumbersCount));
-                        if (difference > -1.001 && difference < 1.001) {
+                        if (testDecimalNumbersCount == goldDecimalNumbersCount && difference > -1.001 && difference < 1.001) {
                             continue;
                         }
                     }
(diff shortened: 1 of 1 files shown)


Tests/GoldEtlCsvTests.cpp
```diff
@@ -102,21 +102,42 @@ class Tests : public ::testing::Test, TestArgs {
...
+
+                    // Floating point may be inconsistently rounded by printf accross different platforms.
+                    // Do a rounding check by ensuring the difference between the two numebrs are less than 
+                    // the final digit +-1.001
+
+                    // Doubles should be good for 15 digits of precision.
+
+                    double testNumber = strtod(a, nullptr);
+                    double goldNumber = strtod(b, nullptr);
+                    const char* testDecimalAddr = strchr(a, '.');
```
JeffersonMontgomery-Intel, line 117: I think this fails for a==b=="0.000".  You could use sscanf() or something to test if the parse succeeded.

Tests/GoldEtlCsvTests.cpp
```diff
@@ -102,21 +102,42 @@ class Tests : public ::testing::Test, TestArgs {
...
+                    const char* goldDecimalAddr = strchr(b, '.');
+                    if (testNumber != 0.0 && goldNumber != 0.0 && testDecimalAddr != NULL && goldDecimalAddr != NULL) {
+                        size_t testDecimalNumbersCount = (a + strlen(a)) - testDecimalAddr - 1;
+                        size_t goldDecimalNumbersCount = (b + strlen(b)) - goldDecimalAddr - 1;
+                        double difference = (testNumber * pow(10.0, testDecimalNumbersCount)) - (goldNumber * pow(10.0, goldDecimalNumbersCount));
+                        if (difference > -1.001 && difference < 1.001) {
```
JeffersonMontgomery-Intel, line 123: You shouldn't multiply testNumber and goldNumber by different factors here, so check that testDecimalNumbersCount==goldDecimalNumbersCount first.  It may be better to just scale the '1' instead (?).

JeffersonMontgomery-Intel (changes requested):
Good catch, I didn't know this about printf().

Mostly looks good but I think I found a few edge cases.

Tests/GoldEtlCsvTests.cpp
Zu-shi: Yeah, that makes sense. I think scaling the 1 means scaling 1.001/-1.001? Can do that.

Zu-shi: Ended up comparing the decimal magnitude between the two numbers instead. It still has the assumption that the spreadsheet will have the same number of decimal places, but I think that's an reasonable expectation and the code is slightly more intuitive if the multiplication happens to the scanned numbers rather than 1.001.

PR #100, opened 2021-03-26
Fix presentmon mpo classification

Zu-shi:
Fix an assumption that HandleDxgkMMIOFlipMPO event means the present must have MPO.

HandleDxgkMMIOFlipMPO is sent even if there is only one active plane. The V/HSyncDPC events are now used to classify MPO.


jenatali (approved):
LGTM

PR #103, opened 2021-05-10
Remove TokenState::Retired event, which is not reliable on Multi-Mon going forward

Zu-shi:
TokenState::Retired events are no longer guaranteed to be sent at the end of a frame, so we use DWM's present stats to understand the Composed Flip timeline instead.

Signed-off-by: Zuoming Shi <zushi@microsoft.com>


PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1300,9 +1303,11 @@ void PMTraceConsumer::HandleDWMEvent(EVENT_RECORD* pEventRecord)
         if (eventIter != mWin32KPresentHistoryTokens.end()) {
             TRACK_PRESENT_PATH(eventIter->second);
             DebugModifyPresent(*eventIter->second);
-            eventIter->second->DwmNotified = true;
```
jenatali, line 1310: Why'd you delete the setting of `DwmNotified`?

PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -1161,6 +1162,8 @@ void PMTraceConsumer::HandleWin32kEvent(EVENT_RECORD* pEventRecord)
         {
             TRACK_PRESENT_PATH(eventIter->second);
 
+            event.SeenInFrameEvent = true;
```
jenatali, line 1165: I thought the ordering here was `ScheduleCompositionSurfaceUpdate` first, and then `InFrame` afterwards. Is it inverted?

Zu-shi: Yeah, InFrame comes first. From test_case_1.etl
image

(A second image to fit the Composition Surface Luid in)
image

PresentData/PresentMonTraceConsumer.cpp
Zu-shi: Ah, that was a typo, fixed.

PresentData/PresentMonTraceConsumer.cpp
jenatali: Huh, fair enough.

jenatali (approved):
Looks okay to me now.

jenatali (approved):
Makes sense to me.

PR #104, opened 2021-05-21
Fix virtual keycode parsing of alphabet keys for -hotkey argument

aslatas:
There is an off-by-one issue in the parsing of the -hotkey argument for alphabet keys (numeric, function, and modifier keys are not affected). For example, when using the command line argument "-hotkey A", the A key does not start a recording. Instead, the B key does.

I think this is just a copying error in the lookup table for the VK codes. See the microsoft docs [here](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes). The range for alphabet keys should start with 0x41 for 'A', but the command line parser interprets 0x42 (the 'B' key) as 'A'. The values for all of the alphabet keys have the same issue, and the 'Z' key isn't present.

This change just updates the lookup table in CommandLine.cpp to use the correct virtual key codes for each letter.

Signed-off-by: Matthew McLaurin<matthew@matthewmclaurin.com>


JeffersonMontgomery-Intel:
Wow, thanks!  Not sure how that happened :)

PR #107, opened 2021-08-25
Fix IFlip Notify Mode where presents could be incorrectly dropped

Zu-shi:
After some recent changes to DWM, some TokenStateChanged events can now come much later than expected. This causes iFlip presents that previously wait on that event to incorrectly become marked as dropped when a newer frame completes first.

We resolve this by allows iFlip presents to be completed without waiting on TokenStateChanged events (e.g. they can be completed by VSyncDPCMPO events). For iFlip, we don't need TokenStateChanged events to track the data we need.

This was locally validated to no longer falsely report drops on a trace NVidia submitted, and passes existing Gold Standard tests.


Zu-shi:
Pinged NVidia a few times on the fix bug didn't get a response. I think this PR is safe to complete based on the issue no longer reproing in the trace they gave us.

PR #111, opened 2022-05-02
Removed Microsoft_Windows_DxgKrnl::Keyword::Microsoft_Windows_DxgKrnl…

ElysiumPtolemus:
…_Performance from TraceSession.cpp

DxgKrnl Keyword has been having issues with performance in various areas. Removing it as it has no benefit currently.


JeffersonMontgomery-Intel:
anyKeywordMask = 0 makes sense; good catch.

But shouldn't allKeywordMask still have both Microsoft_Windows_DxgKrnl_Performance  and Base?

Looking at the events we consume (https://github.com/GameTechDev/PresentMon/blob/main/PresentData/ETW/Microsoft_Windows_DxgKrnl.h#L95) all have 0x4000000000000001, so they should match while further excluding any events that have Base but not Performance.

ElysiumPtolemus:
Thanks for the info! 

Is there a way to specifically remove all references to the Performance Keyword (should I remove the 0x40000001s) + is there anything else that should be changed? We were curious if there were any possible implications for PresentMon that would happen if we do remove this keyword. I did some builds without, and they seemed to pass the tests fine + don't seem to be referencing Performance as Base is the primary keyword used, but wanted to double check since I haven't really interacted with these keywords before.

JeffersonMontgomery-Intel:
Here is the relevant documentation: https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-enabletraceex2

If I'm reading it correctly, both the MatchAnyKeyword and the MatchAllKeyword (and Level) criteria have to be met for an event to be emitted from the provider, so I don't understand why setting the keyword mask(s) to 0 would help.  My understanding is that would cause the DXGK provider to generate _more_ events into ETW, leading to worse performance.

Can you explain what the performance issue is you're trying to solve?  e.g., are you seeing the DXGK provider emit events that aren't used, or is there some other side effect of using this filter bit?

ElysiumPtolemus:
Just pushed more changes for the removal - noticed that Performance also has a channel of 0x11, and a lot of things seem to use these (e.g. all the Event Descriptors). Is there a way to remove them or replace them with something benign?

JeffersonMontgomery-Intel:
The header files in ETW/ are autogenerated from ETW manifests.  In some cases they're edited manually to add extra information (e.g., win7 event info) but I'd rather not modify the event descriptions coming from the manifest.

If we want to ignore the Performance keyword bit, IMO the right place to do that is when enabling the DXGK provider similar to your original changes.

The only open is: why is using the Performance keyword a problem?

ElysiumPtolemus:
Just sent you a mail with the explanation I was given

JeffersonMontgomery-Intel:
FYI: Branch v1.7.1-noperf also applies this PR to v1.7.1

[PresentMon-1.7.1-noperf-x64.exe.zip](https://github.com/GameTechDev/PresentMon/files/8712305/PresentMon-1.7.1-noperf-x64.exe.zip)
[PresentMon-1.7.1-noperf-x86.exe.zip](https://github.com/GameTechDev/PresentMon/files/8712306/PresentMon-1.7.1-noperf-x86.exe.zip)
.

PR #122, opened 2022-10-07
Fix: NOMINMAX macro definition with #ifndef, using std::max/std::min instead of WINAPI ones

IvanZinkevich:
The reason: I have project wide defined `NOMINMAX` and get macro redefinition warning. I decided to wrap it up with `#ifndef` everywhere and change WinAPI max/min calls with STL functions.


JeffersonMontgomery-Intel:
Merged into main with a few tweaks, sorry for the delay.

PR #167, opened 2023-09-03
Fix incorrect CPU utilization statistics on Windows 11 22H2 and newer

Kaldaien:
### Measures of CPU Utilization are fundamentally broken in the latest release of Windows 11.

Somewhere leading up to the release of Windows 11 22H2, Microsoft broke the accumulation of time in `SystemProcessorPerformanceInformation`. Idle CPU time no longer has the same relation to user/kernel/interrupt and the times reported cannot be used to measure CPU load because idle time is orders of magnitude larger than it should be.

Most monitoring software such as Special K, MSI Afterburner and HWiNFO have worked-around the problem by sourcing idle CPU time from `SystemProcessorIdleInformation` and then measuring that against the non-idle counters in `SystemProcessorPerformanceInformation`. Load percentages are consistent with pre-22H2 versions of Windows when using this alternate measure of idle time.

<hr>

While PresentMon does not directly interface with those structs or `NtQuerySystemInformation`, it seems the WMI "Processor" data provider does and is subject to the very same bug the rest of us writing performance monitoring software have worked around.

I have confirmed on my copy of Windows 11 23531.1001 that sampling `\Processor(_Total)\% Processor Time` behaves the same as computing busy vs. (incorrectly accumulated) idle time from `SystemProcessorPerformanceInformation`.

Under the heaviest CPU loads possible, WMI's bungled measure of `% Processor Time` rarely makes it above **7-8%**. Meanwhile, `\Processor(_Total)\% Idle Time` scales as expected in response to load.

<hr>

Manually computing utilization as the inverse of `% Idle Time` produces expected load percentages on Windows 11. On older versions, the inverse of `% Idle Time` is close enough to `% Processor Time` not to matter, so we might as well apply the workaround on all versions of Windows.


planetchili:
Thank you for the contribution. This change has already been made in the interim, essentially has you had proposed it. Apologies for not reviewing the PR in a timely fashion.

PR #192, opened 2023-12-12
Enhance UNICODE support

### LINKED ISSUE #191: Crash when launching an application with non-ANSI characters in filename
When launching a process with non-ANSI characters in filename (e.g. try to copy notepad.exe and rename it to include non-ANSI characters), PresentMon crashes in HandleProcessEvent here: https://github.com/GameTechDev/PresentMon/blob/b4695462b4f10d681a0aabc243f7aa10d5b59eff/PresentData/PresentMonTraceConsumer.cpp#L2195

Problem is that if wcstombs_s fails to convert the process name to ANSI, the resulting string size will be 0.
Best solution would be to change ImageFileName to wstring rather than a plain char string. This will also allow proper filtering of process names with non-ANSI chars.

malikm:
Also to fix #191

### DIFF
--- a/PresentData/PresentMonTraceConsumer.cpp
+++ b/PresentData/PresentMonTraceConsumer.cpp
@@ -2215,7 +2215,8 @@ void PMTraceConsumer::HandleProcessEvent(EVENT_RECORD* pEventRecord)
             };
             mMetadata.GetEventData(pEventRecord, desc, _countof(desc));
             event.ProcessId     = desc[0].GetData<uint32_t>();
-            event.ImageFileName = desc[1].GetData<std::wstring>();
+            std::string str     = desc[1].GetData<std::string>();
+            event.ImageFileName = std::wstring(str.begin(), str.end());
             event.IsStartEvent  = true;
         } else if (hdr.EventDescriptor.Opcode == EVENT_TRACE_TYPE_END||
                    hdr.EventDescriptor.Opcode == EVENT_TRACE_TYPE_DC_END) {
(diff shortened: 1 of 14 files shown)


PresentData/PresentMonTraceConsumer.cpp
```diff
@@ -2218,7 +2215,7 @@ void PMTraceConsumer::HandleProcessEvent(EVENT_RECORD* pEventRecord)
             };
             mMetadata.GetEventData(pEventRecord, desc, _countof(desc));
             event.ProcessId     = desc[0].GetData<uint32_t>();
-            event.ImageFileName = desc[1].GetData<std::string>();
+            event.ImageFileName = desc[1].GetData<std::wstring>();
             event.IsStartEvent  = true;
```
JeffersonMontgomery-Intel, line 2219: AFAIK the underlying data for this event is always char, and GetData<T>() doesn't do any conversion, it just accesses the event data as if it were a T.  So, we need to leave it as GetData<string>() and then convert here.

In debug build GetData() should be asserting with this change, though you'll only hit this event when processing from an ETL file (--etl_file).

JeffersonMontgomery-Intel (changes requested):
Thank-you for the contribution!

There is one change that is needed (see inline comment).  I probably would have gone with wchar_t vs WCHAR, but don't bother changing that now.

I will pull, test, and accept this sometime this week.  I can fix that one event then so you don't need to, this is just FYI.

malikm:
Thanks for the review. I did all changes requested, hope it's OK now :)

JeffersonMontgomery-Intel:
Thanks!  Testing now and everything looks good.  I plumbed it through to the CSV too so should be supported throughout now:

image

Should land in main branch today or tomorrow.

PR #205, opened 2024-03-08
Fix memory leak

malikm:
A node in HwQueueContext allocated via "new" is not automatically de-allocated.


planetchili:
This issue was certainly valid, but has been addressed separately. Apologies for not reviewing your PR in a timely fashion and thank you for the contribution.

PR #319, opened 2024-07-12
Backpressure and other options for PresentData debugging

### LINKED ISSUE #312: FrameId and ScreenTime (DisplayTime) CSV columns for debugging
Time in ms (matching timestamps from the verbose trace)
controllable via CLI options (not advertised)

### LINKED ISSUE #313: Backpressure for offline ETL processing
Use condition variable to block processing thread, signal from output thread
Will automatically activate for offline processing only (always disabled for realtime trace)
CLI option (hidden) to override behavior and disable backpressure
Need to handle the corner case of the buffer filling with non-ready presents (in that case we need to not block and overflow presents, otherwise we will deadlock both threads).

planetchili:
backpressure
frame id
screen time absolute

closes #313 
closes #312

### DIFF
--- a/PresentMon/CsvOutput.cpp
+++ b/PresentMon/CsvOutput.cpp
@@ -239,11 +239,11 @@ void WriteCsvRow<FrameMetrics1>(
         break;
     }
     if (args.mWriteDisplayTime) {
-        if (p.ScreenTime == 0.f) {
-            fwprintf(fp, L",%.*lf", DBL_DIG - 1, 0.001 * pmSession.TimestampToMilliSeconds(p.ScreenTime));
+        if (p.ScreenTime == 0) {
+            fwprintf(fp, L",NA");
         }
         else {
-            fwprintf(fp, L",NA");
+            fwprintf(fp, L",%.*lf", DBL_DIG - 1, 0.001 * pmSession.TimestampToMilliSeconds(p.ScreenTime));
         }
     }
     if (args.mWriteFrameId) {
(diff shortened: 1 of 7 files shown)


planetchili:
closes #313
closes #312

PresentMon/CsvOutput.cpp
```diff
@@ -232,6 +238,17 @@ void WriteCsvRow<FrameMetrics1>(
         fwprintf(fp, L",%.*lf", DBL_DIG - 1, 0.001 * pmSession.TimestampDeltaToMilliSeconds(p.PresentStartTime));
         break;
     }
+    if (args.mWriteDisplayTime) {
```
markgalvan-intel, line 241: Two things on this one. 1. I think it's backwards. When ScreenTime is 0 then we want NA. 2. PresentEvent.ScreenTime is a uint64_t so I don't think we should compare with "0.f".

markgalvan-intel: Oh do it like below. :)

markgalvan-intel (changes requested):
Looks good. I just have the one change request in CsvOutput. Let me know if I am off in the weeds.

PresentMon/CsvOutput.cpp
planetchili: good catch, this is for the old metrics format so I catch that when I checked the running output!

markgalvan-intel (approved):
Looks great!

PR #324, opened 2024-07-16
Service as Child

### LINKED ISSUE #323: AppCef Service-as-Child mode
Add option for AppCef to launch its own presentmon service as child in console app mode and connect to it automatically.

Add CLI options to appcef for:

1. Enabling service-as-child mode
2. forwarding pipe/shm/etw names to the service to use, with suitable defaults

This should greatly aid in debugging, especially when paired with VS child process debugging extension.

Additionally, plan to give the Telemetry providers/adapters custom event type for failure due to that subsystem not being available on the system. This will make it possible to filter those exceptions out so they don't cause breakpoints when break-on-exception is enabled.

planetchili:
closes #323


markgalvan-intel (approved):
Neat stuff!

PR #325, opened 2024-07-16
pmOpenSession_ C linkage

### LINKED ISSUE #264: Add C-API interface endpoint for connecting to service with custom pipe-name/introspection-shm-name
### Auto-migrated from intel-sandbox/applications.graphics.presentmonservice/issues/434
<sup>Original creation by planetchili on 2024-05-30 08:03:57+00:00</sup>

Eventually this will make it into the public API, but for the time being we will keep it in internal since the identifiers needed will likely change after IPC system is upgraded. Currently we have a function in internal that does this, but the C++ linkage is problematic for internal customer. Add an additional C-linkage endpoint that just calls into the C++ one.

planetchili:
closes #264

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,7 @@ PresentMon/PresentMon.args.json
 PresentData/PresentData.args.json
 
 IntelPresentMon/SampleClient/log.txt
+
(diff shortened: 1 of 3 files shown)


markgalvan-intel (approved):
Good stuff.

PR #328, opened 2024-07-18
Sa updates

markgalvan-intel:
Update the security attributes for when the service is launched as an application.


planetchili (approved):
already tested and reviewed before the pr!
looks good, and looks like we can use these settings even when running as child launched by appcef

PR #336, opened 2024-08-02
Bug/multi gpu query

### LINKED ISSUE #256: Selecting adapter NVIDIA GeForce GTX 970, overlay only reports "basic" telemetry (dual dGPU desktop)
I have a dual GPU desktop.
I can capture just fine when I explicitly select the _Radeon RX 580 Series_ device, for any set of metrics (regardless of which GPU is being targeted by the application)

If I switch to capture from the _NVIDIA GeForce GTX 970_ it **appears** I can only capture the "basic" preset.
Switching to any other capture profile at runtime has no effect (basic capture continues to be rendered in the overlay).

If I attempt to restart a capture in any mode other than "basic", no overlay is presented _until_ I revert to "basic"
No change between windowed mode or overlay mode

**However**, a capture CSV report seems to suggest it is capturing, it's just the overlay that is not presented
[pmcap-DX12_fullscreen_portrait_repro.exe-240621-094923.csv](https://github.com/user-attachments/files/15921179/pmcap-DX12_fullscreen_portrait_repro.exe-240621-094923.csv)
[pmcap-DX12_fullscreen_portrait_repro.exe-240621-094923-stats.csv](https://github.com/user-attachments/files/15921180/pmcap-DX12_fullscreen_portrait_repro.exe-240621-094923-stats.csv)

Side question: Is it expected that I should be able to capture using the AMD device, for an application targeting the nVidia device? And should such metrics be considered reliable? (this seems to work, regardless of which device is considered the 'primary' desktop device)

DxDiag report attached if at all helpful
[DxDiag.txt](https://github.com/user-attachments/files/15921155/DxDiag.txt)

planetchili:
closes #256

### DIFF
--- a/IntelPresentMon/AppCef/Web/src/App.vue
+++ b/IntelPresentMon/AppCef/Web/src/App.vue
@@ -225,18 +225,20 @@ export default Vue.extend({
         widget.metrics = widget.metrics.filter(widgetMetric => {
           const metric = Introspection.metrics.find(m => m.id === widgetMetric.metric.metricId);
           if (metric === undefined || metric.availableDeviceIds.length === 0) {
-            // If the metric is undefined, this widgetMetric will be removed, so return false
+            // If the metric is undefined, this widgetMetric will be dropped
             return false;
           }
           // If the metric is found, set up the deviceId and desiredUnitId as needed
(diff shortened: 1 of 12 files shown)


markgalvan-intel (approved):
Good stuff.

PR #338, opened 2024-08-07
Add fflush to ConsoleApplication 2.0 metrics csv

### LINKED ISSUE #232: No stdout flush in console PresentMon 2.0. Is it intended?
Noticed issues with much increased data lag in scenarios when I try to read data via stdout and console PresentMon-2.0.0-x64.exe comparing to the previous console versions. I started investigating it and nailed it down to fflush(stdout), which is performed inside UpdateCsv in V1.x but missing in V2.0. Was it intended, or flushing was simply lost during refactoring? If so, can it be returned back please?

In meanwhile I solved it by building a custom version PresentMon-2.0.0-x64.exe and adding the following to the very end of UpdateCsvT:

    if (args.mCSVOutput == CSVOutput::Stdout)
        fflush(stdout);

markgalvan-intel:
Flushes were accidentally removed at some point.  v2.0.1 added them back in, but only for the --use_v1_metrics CSV.  This change adds them back in for the default CSV as well.

Fixes #232

### DIFF
--- a/PresentMon/CsvOutput.cpp
+++ b/PresentMon/CsvOutput.cpp
@@ -307,6 +307,10 @@ void WriteCsvHeader<FrameMetrics>(FILE* fp)
         fwprintf(fp, L",FrameId");
     }
     fwprintf(fp, L"\n");
+
+    if (args.mCSVOutput == CSVOutput::Stdout) {
+        fflush(fp);
+    }
 }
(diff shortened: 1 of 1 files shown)


planetchili (approved):
Looks straightforward enough to me, good stuff!

PR #341, opened 2024-08-14
Feature/asio svc control

### LINKED ISSUE #337: Upgrading service named pipe control channel
Current command channel suffers from the following flaws:

- Adding a new command requires editing many location of code across multiple source files and projects. This makes it more difficult to add new ones.
- The implementation for an action is spread across the source making it more difficult to understand.
- The payloads for actions and responses must be manually serialized with byte level operations, which are error-prone and time consuming to write
- The serialization and deserialization functions are written separately and must be carefully kept in sync to avoid errors
- Current serialization does not support dynamic data, meaning we have fixed upper limits for array/string sizes
- The overlapped transfer operations are multi-step and require a state machine to manage them throughout their lifetime
- The data types for the actions are not cohesive with the actions themselves, and their usage is not uniform (some actions have bespoke types, others use a common type, and the common type contains fields irrelevant to some of the actions that use it).
- The system does not support communication from the service to the clients (neither retained actions with response nor events)
- The system does not inherently track which clients are connected, beyond the pool of sockets (though client tracking is incidentally done as part of refcounting the tracked processes). Tracking this info explicitly in the control action system, along with session data such as metrics consumed, settings requested, etc. will be necessary in the future.
- The system is inherently synchronous, does not readily support asynchronous execution of a lengthy operation.

The upgrade will implement the following:
- Create a single-file action system where an action and it's serialization is defined within a single file that automatically get registered with the service
- Utilize the serialization library Cereal to automate the process of serializing Params and Response structs will full support for std types like string and vector
- Utilize Cereal to allow for single definition that controls both serialization and deserialization, ensuring that they are always kept and sync
- Utilize boost.asio to control the execution of multiple strands of asynchronous connections, simplifying backend code
- Utilize c++20 coroutine support on boost.asio to handle multi-step asynchronous operations without the need for state machines
- Create a symmetric system for communication that will allow future communication from service to clients in the form of retained action and events
- Implement protocol versioning that can handle additional data in the packets which is ignored by older client

Up until this point there are only 5 actions implemented, meaning that the current implementation is entirely sufficient. Upcoming features will require more communication between service and middleware. We would like to get this long-discussed upgrade in before we increase the number of command actions to save effort.

Supports #317 #318 #280 etc.

planetchili:
Enhancement using cereal and boost.asio to make it easier and less error-prone to add/modify server action RPCs. Also lays the groundwork for full bidirectional async communication between client and service, event broadcasts from service, and connection-oriented client sessions.

Logging IPC has also be upgraded to use the common ASIO-based named pipe component unifying the implementations.

closes #337


markgalvan-intel (approved):
Wow. That was dense. I understand the actions but there is a bunch of stuff that will take more time to completely grock.

PR #346, opened 2024-08-29
Eliminate 1000ms ETW latency (true realtime overlay)

### LINKED ISSUE #317: Improve Realtime Graph Latency
Currently the monitoring/graphing of metrics and frame event data (typically via the overlay) has a hard limit of 1 second of latency. This is the because ETW batches events together, invoking the callback in bursts at an interval of 1 sec.

Understandably, the user experience suffers because of this, and it also complicates some aspects of metric processing.

It appears to be possible to manually flush the trace buffers with by using a FLUSH control command. Therefore, it should be possible to achieve latencies order(s) of magnitude lower than we currently have.

### LINKED ISSUE #340: Random crash after a long idle
Version 2.1.1
After idling, the program will crash randomly. After idling for a long time (6-12 hours), the program will experience random crashes. The crash might occur in this code segment. I installed Presentmon through VCPKG, so I can't see the exact calling stack. After importing the CPP file, it shows up here.
image

here is the calling Stack
image

It may need to enter a long-term lock screen state or sleep mode, but I'm not too sure about this, but i have do this all the time

### LINKED ISSUE #344: Fix shutdown behavior of asio pipe server
Destroying the io_context while there are pending operations is causing issues, and this needs to be addressed.
supports #317

planetchili:
Also includes fixes for the Asio-based actions (RPC) system, IPC log connection to service, and race condition fix in service
closes #340 
closes #344 
closes #317


IntelPresentMon/CommonUtilities/Qpc.cpp
```diff
@@ -0,0 +1,65 @@
...
+		}
+		return 1.0 / double(freq.QuadPart);
+	}
+	void SpinWaitUntilTimestamp(uint64_t timestamp) noexcept
+	{
+		while (GetCurrentTimestamp() < timestamp) {
+			std::this_thread::yield();
+		}
+	}
+	double TimestampDeltaToSeconds(uint64_t start, uint64_t end, double period) noexcept
```
markgalvan-intel, line 31: Are you always guaranteed to have end be greater than start?

planetchili: Yes, but it would be a good idea to foolproof this for future usage / changes!

PR #347, opened 2024-08-29
Feature/input update

### LINKED ISSUE #315: Filter out keyboard input in Click To Photon metric

markgalvan-intel:
Added new metric AllInputToPhotonLatency which tracks all device input to photon. In addition, refined ClickToPhotonLatency to only track mouse clicks.

Closes #315


planetchili (approved):
Looks great! Couple of notes:
1) when building I get an unchecked readme file (probably re-generated automatically from the metrics list). Should probably commit that change
2) Do we ever remove entries from mReceivedMouseClickByHwnd? That could constitute an (admittedly very slow) resource leak

PR #349, opened 2024-09-03
Fix Win11 bug found on Witcher3

markgalvan-intel:
Fixed Win11 issue where multiple retrieve message events were being received and this was causing us to not correctly apply the xform message. Also fine tuned the mouse click threshold.


planetchili (approved):
seems good!

PR #361, opened 2024-10-10
Feature/registry

### LINKED ISSUE #345: Error for PM2 UI client connecting to service after reboot
> I’m on Windows 11.
> It seems that the error only shows up the first time opening the app after reboot.  The service is present and running before I see this error.
> If the app is started as Administrator the first time, there is no error.
> Maybe just add a note to the readme, that if this error is shown, try to close and re-open the app or run in administrator mode?

Possible issue with introspection connection, perhaps relating to initial capture of telemetry caps (although not likely in light of the behavior only appearing at system reboot, not after service restart).

### LINKED ISSUE #357: Windows Registry Common Utility
Add a common utility based on winreg to persist/load configuration to/from the windows registry. This will make it easier to configure logging from the service during debug, and is a first step in enabling configuration of the service from the client.

### LINKED ISSUE #358: Control Logging from Registry
Logging settings (level and directory) to be configurable from registry. Cascading implementation prefers cli => registry => default.

planetchili:
Added ability to configure service logging with registry, together with common utility for registry access. This was adding during a bugfix process, so that fix is included here as well (init failure on reboot bug).

closes #357 
closes #358 
closes #345


IntelPresentMon/PresentMonService/CliOptions.h
```diff
@@ -26,8 +26,8 @@ namespace clio
 	private: Group gl_{ this, "Logging", "Control logging behavior" }; public:
 		Option<std::string> logDir{ this, "--log-dir", "", "Enable logging to a file in the specified directory" };
 		Option<std::string> logPipeName{ this, "--log-pipe-name", pmon::gid::defaultLogPipeBaseName, "Name of the pipe to connect to for log IPC" };
-		Flag disableStdioLog{ this, "--disable-stdio-log", "Disable logging to stderr" };
-		Flag disableDebuggerLog{ this, "--disable-debugger-log", "Disable logging to system debugger" };
+		Flag enableStdioLog{ this, "--enable-stdio-log", "Disable logging to stderr" };
```
markgalvan-intel, line 29: I am confused by the command line descriptions. The arg says enable and the description says disable. By enabling the option are you just supposed to know you are disabling something?

PresentMon.props
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Label="UserMacros">
-    <!-- PresentMonVersion should be "dev" on main branch, and the version number (e.g., 2.1.0) on release branches. -->
-    <PresentMonVersion>dev</PresentMonVersion>
+    <!-- PresentMonVersion should be "dev" on main branch, and the version number (e.g., 2.2.0) on release branches. -->
```
markgalvan-intel, line 4: Let's talk about this one. It's a philosophical one. I'm okay changing it but lets just discuss.

PR #386, opened 2025-01-08
Feature/merge candidate omni

### LINKED ISSUE #359: Non-deterministic CSV output in debug verbose mode
There appears to be another non-deterministic mechanism besides the producer-consumer overflow of present events. This one seems not to be related to backpressure. Currently investigating any other race condition-driven mechanisms.

### LINKED ISSUE #360: ETW Trim/Prune Utility
ETW files are often prohibitively large. The sheer number of events means that verbose tracing can take 30 minutes+ per run. Solutions:
1) filter events to the same set of events that presentmon itself filters to when doing realtime tracing
2) Trim events by timestamp to a shorter region of interest, decreasing the verbose trace time proportionally

Challenges:
Set of trace filters is tightly coupled to PresentData logic for starting an actual trace; this needs to be decoupled.
There are events from providers outside the filter set that are received and processed, so these need be identified.
When trimming, care must be taken to maintain the state context that tracking depends on (initial process / context state dumps, etc.)

Filtering works and results in a 97% reduction of file size.
Currently, trimming works except for GPU work tracing. In order to identify the missing events, verbose processing is required, but currently verbose is acting non-deterministically, so that issue must be addressed first. #359

### LINKED ISSUE #377: Automatic struct dumping utility
It will be useful for extensive logging of API interactions etc. to be able to automatically serialize / stringify static structures. Since current C++ reflection tricks tend to fail on structs with a large number of elements, augment with a code generation tool based on Clang that generates dumping (or other) code. Also improve the existing TMP dump utility to recurse into structures and work on ranges and enums.

### LINKED ISSUE #378: Extensive verbose logging of IGCL Telemetry module
Given the challenges of supporting releases with IGCL, we need to log all API interactions so we can triage issues quickly and accurately. Verbose logging for all data flowing through this module will be added.

- [x] Verbose log dumping of all IGCL return values / structs leveraging Clang-generated reflection code
- [x] Verbose log dumping of all insertions to the sample history and all nearest-neighbor lookup/retrievals
- [x] Add the adapter name to all verbose logging to easily differentiate in multi-adapter scenarios
- [x] Log cap bits on every retrieval

planetchili:
closes #377 
closes #378 
closes #359 
closes #360


IntelPresentMon/ControlLib/IntelPowerTelemetryAdapter.cpp
```diff
@@ -22,19 +28,43 @@ namespace pwr::intel
...
 
-        if (auto result = EnumerateMemoryModules(); result != CTL_RESULT_SUCCESS) {
-            throw std::runtime_error{ "Failed to enumerate memory modules" };
+        // check for alchemist (used to enable features whose support not reported by IGCL)
+        // use device name that match Arc followed by A### part number pattern
+        isAlchemist = std::regex_search(GetName(), std::regex{ R"(Arc.*A\d{3})" });
+        pmlog_verb(v::gpu)("Detecting Alchemist").pmwatch(GetName()).pmwatch(isAlchemist);
+
+        // errors are reported inside this function
+        // do not hard-fail on memory module issues, so no need to check error return code
```
markgalvan-intel, line 43: This comment confuses me. It says to check the error return code but it doesn't.

markgalvan-intel (approved):
Looks great. Just the one question.

PR #392, opened 2025-01-15
Set deferral time limit for IPM

markgalvan-intel:
The deferral time limit was not being set for real time captures which was causing deferred presents to incorrectly be classified as lost when cleaning up "stale deferred frames" in CompletePresent.


planetchili (approved):
looks good

PR #411, opened 2025-02-13
API backwards compatibility with dynamic loading + loader helper DLL

### LINKED ISSUE #398: Add DLL Loader Layer to Support Backward Compatibility
By shipping the middleware DLL with the service instead of the client, we ensure that middleware-service compatibility is always guaranteed. Middleware-client compatibility is already guaranteed via C-API contract (this will be hardened/made absolute).

This change requires that clients locate the DLL at runtime and load it dynamically. To facilitate this process, we add a thin Loader Layer DLL that clients can link to with a convenient import .lib, which will transparently locate the middleware, load it, resolve all the function addresses, and forward the API calls.

- [x] POC of loader layer DLL
- [x] Ability to resolve mangled C++ endpoints for internal usage
- [x] Update installer to ship middleware with service, provide loader in SDK
- [x] Implement error checking with new PM_STATUS codes for DLL issues
- [x] add version check mechanism in C-API, loader checks during loading
- [x] Make loading process thread-safe
- [x] Review and implement DLL Main hooks as necessary
- [x] Clean out unused classes / projects (MockMiddleware etc.)
- [x] Repair and prune API2 tests
- [x] Cleanup organization of loader and solution project dependencies
- [x] Update documentation

planetchili:
closes #398


markgalvan-intel (approved):
Looks fantastic.

PR #416, opened 2025-02-26
Feature/igcl percent metrics

planetchili:
New metric additions--mostly percent metrics correlating with metrics we already have.


markgalvan-intel (approved):
These look great!

PR #425, opened 2025-03-12
allow specifying middleware path in sample client & better wrapper errs

planetchili:
enabling sample client to be used with presentmon servers other than the default globally installed service after the new loader update


markgalvan-intel (approved):
Love it.

PR #435, opened 2025-04-10
Fix frames that are marked as discarded incorrectly

nikkumaar:
The way PresentMon marks the completion of frames is to first mark frames as 'Presented' when it receives MMIO MPO flip call and then once Win32k providers’ (running within DWM process) TokenStateChange(TokenState::Discarded) event is received it is marked as completed.

In certain cases, like when there are short bursts of presents, we get multiple back-to-back flips and token tracking thread ends up marking the first frame in the burst as dropped. To fix this issue, we mark the frame as discarded only if the frame already doesn’t have valid ScreenTime.

ETL file: https://drive.google.com/file/d/1uVnWiT9An-p5_BanYQB8e_-FCPd61qXP/view?usp=sharing


planetchili:
Looks great, thanks for the feedback and fix!

PR #440, opened 2025-04-21
Add support of NVIDIA DLSS4 flip-metering

HugoNVDA:
1. Adds a custom build step in PresentMon.vcxproj to compile and embed the manifest of NV DisplayDriver (DD) ETW event in the PresentMon binary.

The custom build step is currently added for x64 platform.

2. Adds NVTraceConsumer for handling NV DD events that informs the metering of each flip.

3. Updates the CSV output to take into account flip metering. Metrics "DisplayedTime" and "MsBetweenDisplayChange" (V1 metrics) now reflect actual flip intervals when DLSS4 flip-metering is engaged.


PresentMon/OutputThread.cpp
```diff
@@ -250,11 +251,33 @@ static void UpdateChain(
...
+        // If chain->mLastDisplayedScreenTime that is adjusted by flipDelay is larger than screenTime,
+        // it implies the last displayed present is a collapsed present, or a runt frame.
+        // So we adjust the screenTime and flipDelay of screenTime,
+        // effectively making screenTime equals to chain->mLastDisplayedScreenTime.
+
+        // Cast away constness of p to adjust the screenTime and flipDelay.
+        PresentEvent* pp = const_cast<PresentEvent*>(p.get());
+        if (!pp->Displayed.empty()) {
+            pp->FlipDelay += chain->mLastDisplayedScreenTime - screenTime;
+            pp->Displayed[0].second = chain->mLastDisplayedScreenTime;
```
markgalvan-intel, line 276: I'm curious if this should also be updating screenTime as well as "pp->Displayed[0].second" since all metric calculations in ReportMetrics1 use screenTime after the call to "AdjustScreenTimeForCollapsedPresentNV1.

markgalvan-intel (commented):
Code looks great. Please take a look at the comment I added when you have a moment.

PresentMon/OutputThread.cpp
HugoNVDA: Good catch.
Indeed AdjustScreenTimeForCollapsedPresentNV1 takes a reference to screenTime so it can update its value, if needed.
I've updated the PR.

PR #446, opened 2025-05-06
Updating CB Size

markgalvan-intel:
Current CB size starts dropping frames at greater than 700 fps. Increasing size to accommodate framerates up to 1000 FPS.


planetchili (approved):
looks good!

PR #452, opened 2025-05-19
Feature/auto cef redist wix

### LINKED ISSUE #450: CEF Redistributable Automation
Cef redistributable binaries and resources must be recorded into WiX .wxs files manually, which is error-prone for a number of reasons. In particular, when the list of files changes, this might not be noticed causing incomplete builds to be shipped. Use heat.exe to automatically generate the WiX definitions.

Additionally, it is cumbersome to setup build env or update with new CEF binaries and libraries. And it is possible someone might do this and forget to update with heat. Create a script that A) runs cmake to generate the .sln file B) uses msbuild to build the CEF wrapper library in Release and Debug configurations C) executes the pull-cef.bat script to pull the binaries into the correct places in the project folder and D) runs the heat script which generates the .wxs files for the CEF binaries and resources.

planetchili:
closes #450

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -28,13 +28,15 @@ you only need Visual Studio.  Ignore the other build and source dependency instr
 
 2. Build the Chromium Embedded Framework (CEF)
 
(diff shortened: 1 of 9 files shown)


markgalvan-intel (approved):
Very nice!

PR #453, opened 2025-05-20
Feature/col and metric update

markgalvan-intel:
column and metric updates


planetchili (approved):
massive, built and ran, no hitches here looks great!

PR #454, opened 2025-05-20
Feature/unelevate

### LINKED ISSUE #436: Launch CEF without elevation even when KernelProcess is elevated
Some usage scenarios involving injection might require ipm to be launch as administrator, but even in such cases we would like to avoid running the CEF child process subtree elevated. Should be possible to adjust the token integrity to a lower level, but in order to do this with boost::process we will need to switch to the V2 API. Look into integrity level SID values instead of using explorer.exe to derive the desired token.

planetchili:
Ensure that CEF processes run at medium integrity even when ipm is run elevated
closes #436


markgalvan-intel (approved):
Looks fantastic. Great work.

PR #462, opened 2025-06-04
Added ability to resize present event cb

markgalvan-intel:


planetchili (approved):
Nice being able to control this as necessary!

PR #522, opened 2025-10-14
Lost Present Update

markgalvan-intel:
Fix to handle infinite complete/lost present loop. Also fix for exception when present start time comes AFTER terminate process event


PresentMon/OutputThread.cpp
planetchili: I wonder here, if we skip processing the termination, when would it eventually get processed?

Issue #524, opened 2025-10-17
ETL Logging via Service/API


planetchili:
Etl logging functionality added to service/sdk/GUI will allow more effective triage of ETW processing issues, reducing friction in getting users to obtain and run xperf capture scripts and also reducing the size of traces that need to be transferred by users.

PR #526, opened 2025-10-17
Feature/svc etw log

### LINKED ISSUE #467: ETL Trace Collection from Intel PresentMon UI
This will enable much better compliance when additional data is required to diagnose issues reported by users (no need to install extra tools, run complicated scripts, etc.) Will enable us to use the exact same provider set/filters used when actually running presentmon service. Will also be useful for users as a way of recording ETL traces from a GUI.

- [x] Enable the service to run trace session to logfile
- [x] ETL logging sessions should not interact with realtime sessions so both can run concurrently
- [x] Add to the service a set of actions for controlling the trace behavior
- [x] Make sure that trace logfile options do not give users a way of escalating privileges
- [x] Implement PresentMon API endpoints for controlling ETL logging
- [x] Implement PresentMon UI for ETL tracing

planetchili:
Added secure folder utility, logger component to service, and new (rundown only) ETW event to track from kernel-process
closes #467 #524

### DIFF
--- a/IntelPresentMon/Core/source/pmon/DynamicQuery.cpp
+++ b/IntelPresentMon/Core/source/pmon/DynamicQuery.cpp
@@ -16,7 +16,7 @@ namespace p2c::pmon
 				.arrayIndex = qmet.arrayIndex,
 			});
 		}
-		query = session.RegisterDyanamicQuery(elements, winSizeMs, metricOffsetMs);
+		query = session.RegisterDynamicQuery(elements, winSizeMs, metricOffsetMs);
 		blobs = query.MakeBlobContainer(1u);
 	}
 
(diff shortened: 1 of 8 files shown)


IntelPresentMon/CommonUtilities/file/TempFile.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+	TempFile::~TempFile()
+	{
+		if (!Empty()) {
+			fs::remove(path_);
+		}
+	}
+	TempFile TempFile::Create(const std::string& nameIn)
+	{
+		return CreateAt(fs::temp_directory_path());
```
markgalvan-intel: I believe the nameIn parameter should be passed along in the CreateAt function.

IntelPresentMon/Core/source/pmon/DynamicQuery.cpp
```diff
@@ -16,7 +16,7 @@ namespace p2c::pmon
 				.arrayIndex = qmet.arrayIndex,
 			});
 		}
-		query = session.RegisterDyanamicQuery(elements, winSizeMs, metricOffsetMs);
```
markgalvan-intel: Wow! This is a great catch!! Dare I see this is the change of the PR?

markgalvan-intel (approved):

PR #527, opened 2025-10-17
check the enabled field for sustained power limit

### LINKED ISSUE #517: Service incorrectly detects sustained power limit telemetry available for BMG
Appears as option in IPM, but always reports 0.  Expected to not be available on BMG and should report .enabled=false in the telemetry item.

planetchili:
Fix sustained power limit being incorrectly reported as available for BMG
closes #517


markgalvan-intel (approved):
Nice catch!

PR #529, opened 2025-10-20
Prevent pruning of DWM.exe swap chains with address 0x0

markgalvan-intel:
Pruning DWM.exe swap chains with address 0x0 caused inconsistent test results because these swap chains are reused by the Desktop Window Manager for desktop composition. Removing them prematurely led to missing or unexpected data during unit and integration tests, especially in scenarios where DWM activity is infrequent. By excluding DWM.exe swap chains with address 0x0 from pruning, we ensure stable and predictable test outcomes.

### DIFF
--- a/PresentMon/OutputThread.cpp
+++ b/PresentMon/OutputThread.cpp
@@ -4,10 +4,13 @@
 
 #include "PresentMon.hpp"
 #include "../IntelPresentMon/CommonUtilities/Math.h"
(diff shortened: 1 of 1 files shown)


planetchili (approved):
Great stuff! I wonder about a theoretical case where dwm 0x0 chain could sit stale
but then again, it's max 1 stale chain so it's not like it's gonna leak a bunch of memory
added a small cleanup of the string processing code, if it looks good go ahead and merge!

PR #533, opened 2025-10-30
Fix hotkey no csv

DealsBeam:
This pull request fixes a bug where the --hotkey command-line argument was incorrectly disabled when the --no_csv flag was used. This prevented users from using hotkeys to control recording unless a CSV file was being generated.

The fix removes the logic in CommandLine.cpp that erroneously linked hotkey support to CSV output, ensuring that hotkeys function independently.
Verification

To confirm the fix, a new unit test, HotkeyWithoutCsv, has been added to Tests/CommandLineTests.cpp. This test:

    Launches PresentMon with both --no_csv and a --hotkey argument.
    Simulates the specified hotkey press.
    Captures the console output and asserts that the "Started recording" message is present, confirming the hotkey was processed correctly.

This test was added to address the code review feedback and ensure the fix is verifiable.

Note: While the fix and the corresponding test case have been implemented, I was unable to compile and run the test suite due to persistent build environment issues. The changes are logically sound, but will require verification by running the tests in a configured environment.


planetchili:
Thank you for the PR. We have reviewed and come to the conclusion that this is by design. Typically you would need a hotkey when running a full screen game and collecting csv trace data. Without the csv trace there is little motivation to enable the hotkey.

This PR also includes build scripting / CI enhancements that overlap with ones already added recently, and merge conflicts, complicating matters further.

PR #557, opened 2025-12-22
Update TargetBlockList with additional executables

### LINKED ISSUE #461: nvidia overlay.exe not included in targetblocklist.txt
There's almost no reason one would need to monitor the performance of a performance monitor, and since the overlay often launches alongside games, it tends to lead to Presentmon hooking onto the wrong process. This means minimizing/alt-tabbing out of the program, which can make it difficult to gauge if an issue is due to overall performance, the nvidia overlay, or due to entering/exiting full-screen mode.

Shinigami92:
Added new entries to the TargetBlockList.

closes #461

tested on my PC


markgalvan-intel (approved):
Looks good. Thanks for the contribution. Much appreciated.

PR #569, opened 2026-01-18
Update TargetBlockList.txt To exclude many more non graphic processes

loknath2002:
Please consider the PR, Added processes which are not required to be monitored for graphic properties in real-time.
I have tested this on my system. 
The list is bit exhaustive as I have collected this over 2 months on 2 different PC's
This may not fix any open issues however I am hoping that this would make the game detection process a little easier.
Please do let me know if anything needs to be removed. Thanks


Shinigami92 (approved):

planetchili (approved):
looks great, thank you for the contribution!

PR #584, opened 2026-02-10
Feature/new mid

### LINKED ISSUE #275: PresentMon Library
### Auto-migrated from intel-sandbox/applications.graphics.presentmonservice/issues/402
<sup>Original creation by markgalvan-intel on 2024-02-01 00:54:56+00:00</sup>

Library for common calculations

### LINKED ISSUE #284: Upgrade IPC to use Boost.IPC w/ Managed Allocator
### Auto-migrated from intel-sandbox/applications.graphics.presentmonservice/issues/391
<sup>Original creation by planetchili on 2024-01-23 10:50:17+00:00</sup>

- replace named pipes control/command channels with Boost.IPC based system having single-file RPC definition supporting async call
- replace current nsm frame buffer with Boost shared memory-aware circular buffer, modular design freely supporting organization of frame data
- dynamic binding of middleware to nsm data, supporting flexible data layout with varying metric array sizes and devices, and enabling backward / forward compatibility between middleware and service
- async event channel from service => middleware

### LINKED ISSUE #403: Free telemetry from frame data
Having the telemetry data tied to frames in the IPC shared memory has several drawbacks:

- Redundant telemetry usage when frame rate exceeds telemetry sample rate
- Redundancy when tracking multiple processes
- Redundancy multiplied by number of devices in multi-device scenarios
- Inability to measure telemetry when frame rate is low / non-existent
- Extra work aligning telemetry data that may not be consumed

Storing the telemetry samples in its own structure at a different rate than frame rate solves this. This means when alignment is done (during frame data event capture), it will be done on the middleware side on demand.

### LINKED ISSUE #405: Pull Static Data out of per-frame / per-device data in SHM
Right now static data such as process name or memory capacity is duplicated for every frame. API already differentiates between static and dynamic metrics and has (ill-defined) support for filling static data into dynamic (and frame?) queries. There are some ad-hoc efforts at caching some of these values in various places which adds extra confusion and potential for bugs. We need to store statics in a single location in SHM on a per-process or per-device basis as appropriate.

### LINKED ISSUE #543: Middleware Rework
#284 #275 
The above 2 issues entail considerable changes to the middleware. Since the currently middleware is a placeholder created (due to time constraints) back when the V2 API was implemented, this is a good opportunity to rework it to be sustainable and scalable as originally intended.

Goals for rework:

- Share calculation logic between frame and polled systems (as well as OPM) (#275)
- Telemetry alignment middleware side (#284 #403)
- Full support of multi-adapter queries, removal of current caches and shims (#284 #404)
- Shared access and caching of intermediate frame data between frame and dynamic queries
- Robust log reporting of errors, warnings, info w/ verbose dumping of bulk calculated data
- Improved approach to navigating multiple frames for intermediate generation (here or in IPC or both)
- Middleware reports consumed metrics / tracked devices to service
- Errors are handled consistently with clear paths to recovery or failure (no ignored error conditions)
- Remove redundant caching relating to static data etc.
- Control of stat window placement from API via explicit timestamp (#474)
- Restore multi swapchain behavior
- Better architecture of the Middleware code to address the obj size limit issue and improve organization
- Eliminate the defunct virtualization (ConcreteMiddleware)

planetchili:
Unified metric calculation, reworked shared memory IPC for frame and telemetry data, rewritten middleware for all query paths.
closes #403
closes #405 
closes #275 
closes #284 
closes #543

### DIFF
--- a/IntelPresentMon/Interprocess/source/ShmRing.h
+++ b/IntelPresentMon/Interprocess/source/ShmRing.h
@@ -89,7 +89,7 @@ namespace pmon::ipc
 				const size_t serial = nextWriteSerial_;
 				// once we have looped around the vector, we need to maintain a buffer
 				// to help avoid the client reading partially-updated data (data race)
-				return { nextWriteSerial_ - data_.size() + ReadBufferSize, nextWriteSerial_ };
+				return { serial - data_.size() + ReadBufferSize, serial };
 			}
 		}
 		void MarkNextRead(size_t serial) const


IntelPresentMon/Interprocess/source/ShmRing.h
```diff
@@ -0,0 +1,117 @@
...
+		{
+			if (nextWriteSerial_ < data_.size()) {
+				return { 0, nextWriteSerial_ };
+			}
+			else {
+				// lock current next serial to keep the returned range logically consistent
+				const size_t serial = nextWriteSerial_;
+				// once we have looped around the vector, we need to maintain a buffer
+				// to help avoid the client reading partially-updated data (data race)
+				return { nextWriteSerial_ - data_.size() + ReadBufferSize, nextWriteSerial_ };
```
markgalvan-intel: I think we want to use serial here instead of nextWriteSerial_.

PR #587, opened 2026-02-12
Fixes for both ETL and ULT failures.

markgalvan-intel:


planetchili (approved):
built and run, all expected tests passing!

PR #593, opened 2026-02-19
ensure cleanup of AppData/Local folder on uninstall and install

### LINKED ISSUE #591: UI Exits After 1 Second Pause
Sporadic issue observed with IPM capture application where it opens showing blank UI window and then closes after about 1s. Logs show no errors, service and middleware behaving correctly, no errors on kproc exit or anything else. Only appears in installed Release build from my machines. Only appears when installed/run on some machines (2 + 1 desktops exhibit, 3 laptops tried do not).

planetchili:
closes #591


markgalvan-intel (approved):
Looks good.

PR #596, opened 2026-02-20
Fix/loadout mig new middle

### LINKED ISSUE #589: Loadout Migration For Query Strictness
Middleware rewrite included changes to fix consistency of the metric definitions. It also added thorough validation of query registration arguments. Old loadouts that passed with illogical metric/stat combinations etc. now fail. Migration needs to be added so that old loadouts having bad stat settings etc. are corrected.

planetchili:
Migrate json widget loadouts to be compatible with new stricter/more thorough validation in query registration paths.
closes #589


markgalvan-intel (approved):
Good stuff.

PR #600, opened 2026-03-04
Support for printing source and layer ID in output CSV

kpsam2000Intel:
Support for printing source and layer ID in output CSV. Helps with multi-display offline ETL analysis


planetchili:
Closing as implemented by #603

PR #632, opened 2026-05-29
Feature/granular telemetry

### LINKED ISSUE #352: VRAM usage display bug
image

GPU Memory Size Used reads well over my GPU VRAM capacity which is 10GB ( RX 6700 ), I just opened Elden Ring and it says it's using 37GB. I've also gotten numbers above my GPU's dedicated VRAM in a number of other games.  I would expect GPU vram usage to show at most the capacity of my card, even if it's somehow including assets that were swapped to RAM there is no way it's 37GB after I just opened the game.

Tested on Windows 10 on AMD RX 6700 10GB, ryzen 5600. Version 2.1

### LINKED ISSUE #438: ADL Memory Utilization Telemetry
Current endpoint seems to include both shared memory and dedicated VRAM. There is another newer endpoint that seems to report only dedicated VRAM usage. We should test that out on AMD, preferably in a memory-starvation scenario. Also requested to compare metric to the value reported by Windows (task manager?)

### LINKED ISSUE #579: Service Fails To Restart When Client Still Connected
To reproduce: start client session with running installed service, stop service while keeping client/session, then attempt to start service again. Service will fail to start due a shm file object's name already in existence (possibly the introspection shm).

Proposed solution: add salt string to shm name(s) to avoid collisions.

### LINKED ISSUE #606: Remove Legacy Telemetry Shims for Caps and Polling
Old telemetry used static bitset for caps and static struct for polling output. New system uses dynamic maps for both. To minimize the IPC/Middleware rewrite work, Telemetry (ADL/NVAPI+NVML/IGCL/WMI) was not ported over to new system, just shimmed. These components should natively output the new format.

### LINKED ISSUE #607: Fine-Grained Telemetry Providers
Recent upgrade to Telemetry allows for polling only required devices, rather than always polling all devices. However, arch is still monolithic 1 provider per device, always polling for every metric available regardless of demand.

The next step is to increase provider granularity to a per-metric level. This will enable A) only calling necessary endpoints and B) combining/cascading multiple providers to provide a greater total set of available metrics.

Open problem for consideration: how to correlate device across multiple APIs/providers.

### LINKED ISSUE #616: Accurately detect integrated graphics for Intel/AMD and sort to bottom

planetchili:
closes #607 
closes #606 
closes #616 
closes #438
closes #352
closes #579


markgalvan-intel (approved):
Looks great. I love the new direction! Excellent work.

PR #637, opened 2026-06-02
Fix/nvml struct

planetchili:
Undo nvAPI mismatch workaround, make header dependency explicit/order independent (GHCP-suggested edits from granular telemetry PR).


markgalvan-intel (approved):
Looks great!

PR #641, opened 2026-06-05
Fix/stop high precision waits

### LINKED ISSUE #640: CPU utilization relating to high-precision scheduling
ETW traces have been reported to show significant proportion of service utilization attributed to SwitchToThread. Investigation has shown that, at least on the machines examined, SwitchToThread utilization was far less than utilization attributed to waiting on Windows high-resolution timer (184ms vs 4687ms of core slice over 63 seconds).

The main source is the ETW flushing thread, which runs at 8ms period (125 Hz). If we can switch this to using Sleep we can reduce service utilization (already low at 0.4% on test machine) by a significant proportion. It is not important that flush occurs at a precise cadence; it is not a sampling operation.

However, the maximum duration between flushes directly influences latency of frame event delivery to clients. Observation has shown latency increases at double the rate of increase of flush period. So moving from 8ms (current) to roughly 16ms (Sleep(1)) would potentially increase latency by 16 ms. Still, this does not seem to be a dealbreaker.

The next-highest utilization location for precision waiting is telemetry poll loop. Here we are sampling at a fixed cadence and jitter would be undesirable. However, telemetry typically runs at 100 ms period, and so ~15ms max jitter is likely not a dealbreaker either. Also, since this operation runs only at 10 Hz, it may be permissible to keep the high-precision wait for telemetry.

It is also possible to set the system timer precision higher, up to 1ms precision, at the cost of some system overhead for the higher rate of servicing. This could be opt in for users desiring lower-latency ETW and/or high precision in sampling cadence (lower jitter/skew).

Limits for period settings in the API might need to be adjusted to take the reduced resolution into account (allowing user to set 4 ms ETW or telemetry period makes little sense with 15.6 ms timer resolution. Default dynamic querying offset will need to be adjusted higher to account for the increase frame data delay. Testing should be done to find the new sweet spot.

Consider also selecting between high-precision and Sleep waiting modes at runtime based on API user demands. Same goes for setting system timer resolution.

planetchili:
closes #640 

Stop using high precision waiting for interval timer everywhere. Main offenders are ETW flush thread, followed distantly by telemetry poll thread. This change can bring us from 0.4% utilization to under 0.2% utilization on test system.

Using Sleep() means we can be off by as much as 15.6 ms (and can never schedule faster than that rate). Increased overlay window offset to match increased event processing lag (now 150ms). Telemetry will experience 15.6 ms jitter in sample cadence, but 15.6/100 is likely tolerable. Overlay polling client-side also changed, but with manual window placement we can guarantee perfect polling cadence even without timer precision (some jitter for graph update is likely tolerable, especially if we use standard update rate of 10 Hz). Set minimum ETW flush period to 8 ms and minimum telemetry period to 50 ms to match these changes (requesting out of range no longer errors, just output warning to log/diagnostic).

### DIFF
--- a/IntelPresentMon/AppCef/ipm-ui-vue/src/core/preferences.ts
+++ b/IntelPresentMon/AppCef/ipm-ui-vue/src/core/preferences.ts
@@ -81,7 +81,7 @@ export function makeDefaultPreferences(): Preferences {
         telemetrySamplingPeriodMs: 100,
         etwFlushPeriod: 8,
         manualEtwFlush: true,
-        metricsOffset: 80, 
+        metricsOffset: 150, 
         metricsWindow: 1000, 
         overlayPosition: 0, 
         timeRange: 10, 
@@ -140,7 +140,7 @@ export function makeDefaultPreferences(): Preferences {
 
 export const signature: Signature = {
     code: "p2c-cap-pref",
-    version: "0.20.0",
+    version: "0.21.0",
 };
 
 export interface PreferenceFile {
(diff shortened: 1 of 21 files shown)


markgalvan-intel (approved):
Looks fantastic. Great work.

PR #642, opened 2026-06-05
Fix/update apply app data method

markgalvan-intel:
Updating app provider attachment methodology. Instead of finding and attaching to the closest Present Start, will now attach to the Present where the app provided present start resides in the Present Start and Present End interval. Note this only used in the absence of the app frame it.


planetchili (approved):
Looks good to me, great stuff!

PR #651, opened 2026-06-16
Added new D3D12 PSO compilation metrics

gvinals:
This commit adds 3 new metrics related to D3D12 PSO compilation:
- D3D12 PSO Compile Count: D3D12 PSO compilation events started every second.
- D3D12 PSO Compile Time: D3D12 PSO compile time in milliseconds per second (across all CPUs).
- D3D12 PSO Compile Busy: Percent of time with D3D12 PSO compilation taking place.

This commit also adds a new supported provider (Microsoft-Windows-Direct3D12).

Note that: D3D12 PSO compilation events are not tied to a specific present.

This new events are tied to the process itself and are a "new" type of metric (per-process, ETW based but not tied to a present). This commit includes very small changes to support the new metric type.


planetchili:
Superseded by #663

PR #669, opened 2026-07-17
Calculate AnimationError and AnimationTime for generated frames

markgalvan-intel:
Calculate animation metrics per displayed frame by distributing the application simulation interval across every generated and application display step between two application frames. Buffer rows until both animation context and display timing are available, while preserving ingest order for generated and not-displayed frames. Start a new animation timeline when the simulation source changes, omit animation metrics when no valid interval exists, and retain lookahead buffering so displayed duration remains accurate.


planetchili (approved):
Looking great, thanks!

PR #676, opened 2026-08-06
Adding frame type of DLSS as well

kpsam2000Intel:
Adding frame type of DLSS as well


markgalvan-intel (approved):
Thank you for the PR. As coded this will work from the provider side but will not be processed correctly when handling the PresentFrameType event. I will allow the merge and we will add an issue to implement the necessary PresentFrameType event code.

